Issue #3, opened 2014-01-21
Cannot use JSONWebTokenAuthentication and OAuth2Authentication authentication classes at the same time


marccerrato:
When using both authentication classes at the same time, only the first of them in the authentication_classes list will be checked for authenticating a user, whether successfully or raising a AuthenticationFailed exception.

The source of this conflict seems to be that both classes use the same authorization header keyword 'Bearer'.

Possible solutions I see viable, range from using a different authorization header keyword, to validating the token format to distinguish it from the OAuth2 one.

PR #4, opened 2014-01-22
Added tests to validate JWT Auth compatibility with OAuth2

marccerrato:
I've added the tests to validate the compatibility problem between JWT Auth and OAuth2, which I commented on the issue https://github.com/GetBlimp/django-rest-framework-jwt/issues/3.

I have also used `Tox` to make it possible to run the tests with the extra requirement of `django-oauth2-provider` without altering the base requirements of the package.


bharathi7899:
@marccerrato nice dude

PR #23, opened 2014-06-19
Add refresh token feature

alvinchow86:
Added an optional endpoint that can refresh a token. This can be useful for a webapp to try to keep a user logged in without having to re-enter their password constantly. Got the idea from this post https://auth0.com/blog/2014/01/27/ten-things-you-should-know-about-tokens-and-cookies/

Basically, if `JWT_ALLOW_TOKEN_RENEWAL` is True, then an extra field `orig_iat` (stands for original issue at time) is stored on the returned JWT token. You can then pass the token to the `RefreshJSONWebToken` view, and if `orig_iat` hasn't yet expired, you get a token with a refreshed expiration time (but with the same `orig_iat`, so you can only do refreshes up to some time limit, specified in `JWT_TOKEN_RENEWAL_LIMIT`). Getting a brand new token (by passing in username/password) will return a completely new `orig_iat` though.

I also added a new handler, `jwt_get_user_id_from_payload`, that lets you specify how the `user_id` field is stored. Since it's possible for the custom payload handler to change how the user id is stored, it seems to make sense to have a way to retrieve it too.

Feedback welcome (there are some things that could probably be cleaned up). I didn't add tests yet but if there is interest in this PR I will write some (I did write a number of tests in my application that uses this, so I am reasonably confident it works).

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ $ curl -H "Authorization: JWT <your_token>" http://localhost:8000/protected-url/
 ```
 
 ## Refresh Token
-If `JWT_ALLOW_TOKEN_REFRESH` is True, issued tokens can be "refreshed" to obtain a new brand token with renewed expiration time. Add a URL pattern like this:
+If `JWT_ALLOW_REFRESH` is True, issued tokens can be "refreshed" to obtain a new brand token with renewed expiration time. Add a URL pattern like this:
 ```python
     url(r'^api-token-refresh/', 'rest_framework_jwt.views.refresh_jwt_token'),
 ```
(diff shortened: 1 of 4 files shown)


jpadilla:
@alvinchow86 any reason why we used `orig_iat` here instead of `iat`? This seems to might have slipped by.

alvinchow86:
`iat` time is a standard field in the JWT spec and means the time this particular token was issued. I didn't want to override that definition (and it's still useful to know), so I added a new field `orig_iat` to track the `iat of the original token in the refresh chain (it could be renamed anything else of course. is that what you were wondering about?

jpadilla:
@alvinchow86 got it, thought we had incorrectly named it and thought it was meant to just be `iat`. Thanks for the clarification.

PR #59, opened 2015-01-01
Changed potentially misleading error message

skolsuper:
If a custom user model uses a different name for the 'username' field, the error message thrown by JSONWebTokenSerializer.validate in case of a missing field is incorrect.

### DIFF
--- a/rest_framework_jwt/serializers.py
+++ b/rest_framework_jwt/serializers.py
@@ -68,7 +68,10 @@ class JSONWebTokenSerializer(Serializer):
                 msg = 'Unable to login with provided credentials.'
                 raise serializers.ValidationError(msg)
         else:
-            msg = 'Must include "username" and "password"'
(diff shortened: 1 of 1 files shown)


jpadilla:
@skolsuper yeah seems good. Note that the Travis CI build failed.

> rest_framework_jwt/serializers.py:74:17: E123 closing bracket does not match indentation of opening bracket's line

skolsuper:
My bad. Fixed now.

jpadilla:
@skolsuper also I think we can drop the substitution for password and do something like this instead.

```python
msg = 'Must include "{0}" and "password"'.format(self.username_field)
```

skolsuper:
If you want to fit this onto one line I could change it to
```msg = 'Must include "%s" and "password"' % self.username_field```

jpadilla:
@skolsuper Nah don't care about fitting it onto one line. I think it's good as it is.

PR #61, opened 2015-01-05
Adds "JWT_RESPONSE_PAYLOAD_HANDLER"

erichonkanen:
Allows easily attaching additional data to the login response such as the currently logged in User.

*PS I think my working repo was out of date so it includes what was merged from current master

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,9 @@ JWT_AUTH = {
     'JWT_PAYLOAD_GET_USER_ID_HANDLER':
     'rest_framework_jwt.utils.jwt_get_user_id_from_payload_handler',
 
+    'JWT_RESPONSE_PAYLOAD_HANDLER':
+    'rest_framework_jwt.utils.jwt_response_payload_handler',
+
     'JWT_SECRET_KEY': settings.SECRET_KEY,
     'JWT_ALGORITHM': 'HS256',
     'JWT_VERIFY': True,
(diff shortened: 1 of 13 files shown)


jpadilla:
@erichonkanen 659a88e seems to be adding files all over the root of the project.

On another note, I'm not really sure this would actually work at all since [this line](https://github.com/GetBlimp/django-rest-framework-jwt/blob/master/rest_framework_jwt/views.py#L26) is what actually returns the response. I would expect people to just create a custom APIView from `ObtainJSONWebToken`.

If we were to use `jwt_response_payload` on our existing API views, our serializer must return a `user` as well as the already existing `token` in the `validate()` method, and then something like:

```python
data = {'token': serializer.object['token']}
data.update(jwt_response_payload(user))
return Response(data)
```

Wonder how helpful it might be to pass the request as well to `jwt_response_payload`.

erichonkanen:
@jpadilla 

Sorry about the files, what happened is I just did an init in my virtualenv copy and then pulled down my fork which had newer changes than my working copy so it added those to the commit...

Also in the multiple commits it is fixed:

https://github.com/erichonkanen/django-rest-framework-jwt/commit/148daaaac61320a520c0b9cef0db6805b714e242#diff-5afafe00fb1f34cf237d8be61f350e1bR26

```
    return Response(serializer.object)
```

I have this tested and working locally, works nicely. I can resubmit the PR if you want to get cleaner look..  I think this would be really nice feature to have..

erichonkanen:
Ah I see what you mean about code being added, I'll resubmit this!

PR #62, opened 2015-01-06
Added jwt_response_payload_handler which allows returning additional response data

erichonkanen:
Adds the ability to attach additional info besides the "token" to the response for the obtain and refresh endpoints.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,14 +179,16 @@ Specify a custom function to generate the token payload
 If you store `user_id` differently than the default payload handler does, implement this function to fetch `user_id` from the payload.
 
 ### JWT_RESPONSE_PAYLOAD_HANDLER
-Attach additional data to the auth login response such as the currently logged in User. Accepts the User object as argument.
+Responsible for controlling the response data returned after login or refresh. Override to return a custom response such as including the serialized representation of the User.
+
+Defaults to return the JWT token.
 
 Example:
 ```
-def jwt_response_payload(user):
-    serializer = UserSerializer(user)
+def jwt_response_payload_handler(token, user=None):
     return {
-        'user': serializer.data
+        'token': token,
+        'user': UserSerializer(user).data
     }
 ```
 
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ def jwt_response_payload_handler(token, user=None):
     }
 ```
 
-Default is `{}`
+Default is `{'token': token}`
 
 ### JWT_AUTH_HEADER_PREFIX
 You can modify the Authorization header value prefix that is required to be sent together with the token. The default value is `JWT`. This decision was introduced in PR [#4](https://github.com/GetBlimp/django-rest-framework-jwt/pull/4) to allow using both this package and OAuth2 in DRF.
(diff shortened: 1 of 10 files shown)


erichonkanen:
@jpadilla lmk if this looks better... I think it would be nice to only have to implement a custom jwt_response_payload_handler()" function like I put in the README example but if you think this should be left to a custom auth APIView let me know...

jpadilla:
Failing test.
> rest_framework_jwt/serializers.py:65:1: W293 blank line contains whitespace

Also, I think I prefer to have anything related to additional response data handled in the views not in the serializer. So basically serializers stay as they were. Makes sense?

erichonkanen:
@jpadilla ah ok gotcha, do you like the idea of being able to attach additional response data w/o having to write a whole custom auth view though? To me it seems nice but maybe I'm biased... I'll move it to the view and check back later..

erichonkanen:
@jpadilla updated per conversation and cleaned up... Only thing that may be missing is any additional tests, if you have any ideas on that let me know otherwise it should be good to go! Thanks for the assistance

jpadilla:
@erichonkanen I think we should have at least one test that shows custom response data in a view. Everything else looks good to me. Great job!

erichonkanen:
@jpadilla So I ran into some trouble bc the project I'm testing with uses a custom user model and it hiccup'd when I ran test.

That said is this the sort of test you wanted? Or something more involved? Also not sure if theres a better way to condense that decorator override_settings()

```
class ObtainJSONWebTokenTests(BaseTestCase):
    @override_settings(JWT_AUTH['JWT_RESPONSE_PAYLOAD_HANDLER']='tests.utils.jwt_response_payload_handler')
    def test_jwt_login_custom_response_json(self):
        """
        Ensure JWT login view using JSON POST works.
        """
        client = APIClient(enforce_csrf_checks=True)

        response = client.post('/auth-token/', self.data, format='json')

        decoded_payload = utils.jwt_decode_handler(response.data['token'])

        self.assertEqual(response.status_code, status.HTTP_200_OK)
        self.assertEqual(decoded_payload['username'], self.username)
        self.assertEqual(response.data['user'], self.user)
```

jpadilla:
@erichonkanen yeah I think that's the idea.

erichonkanen:
@jpadilla Hey so I pushed up my latest with changes to test to a) fix User model issue, b) include test for custom response. However I can't test this as I'm having issues with testing for some reason.

What are your steps to test this package? I am (in active py3.4 virtualenv) "django-admin test" while in ~/.virtualenvs/project/src/djangorestframework-jwt/"

I am getting operational error saying no CustomUser model (from /tests/models.py).

I can try to resolve this or maybe you can test what I have? LMK.. I probably need to get testing resolved for future reference anyways

jpadilla:
@erichonkanen try running tests with `./runtests.py` instead of `django-admin test`

erichonkanen:
debugging issue that cropped up, the refresh functionality is working but now I'm getting signature has expired.. tricky to track down what's causing this.. seems like the token isn't actually being refreshed

Issue #63, opened 2015-01-09
Per-User Secret Key


DavidMuller:
I've been thinking about the possibility of signing each JWT with a user-specific secret key.  Imagine something like:

```python
def jwt_encode_handler(payload, user):
    return jwt.encode(
        payload,
        user.secret_key,
        api_settings.JWT_ALGORITHM
    ).decode('utf-8')
```

This would make a web application less vulnerable to an attack if  their master `JWT_SECRET_KEY` were somehow leaked--user privacy would not be compromised because each user has an individual key.

Having JWTs signed on a per-user basis, would also conveniently allow a user to "log out of all sessions." (If a user's secret key is updated, then all previous JWTs issued to them are immediately made invalid.) 

I wanted to see what you thought about this; the current implementation would not really adapt its self well to the above idea, but I'd be happy to come up with a PR.

theskumar:
+1 to the idea of having user specific `secret_key` and reasoning behind it.

DavidMuller:
I've taken a stab at this in #65

PR #65, opened 2015-01-09
Add user=None Arguments To Encode and Decode Handlers

DavidMuller:
This PR makes it possible to access a `user` inside  `jwt_encode_handler()` and `jwt_decode_handler()` --> the encode/decode handlers now have signatures like:
```python
def jwt_encode_handler(payload, user=None):
```

I've passed the appropriate `user` object to all instances of `jwt_encode_handler()` in the code base.   

Someone who wanted to take advantage of a `user` object in `jwt_decode_handler()`, however, would need to override `RefreshJSONWebTokenSerializer.validate()` and `JSONWebTokenAuthentication.authenticate()` with logic that would retrieve a user before calling `jwt_decode_handler()`.  I've abstracted functionality in `JSONWebTokenAuthentication` and `RefreshJSONWebTokenSerializer`  to make such overrides easier.

Related issue: #63

### DIFF
--- a/rest_framework_jwt/authentication.py
+++ b/rest_framework_jwt/authentication.py
@@ -30,10 +30,12 @@ class JSONWebTokenAuthentication(BaseAuthentication):
     """
     www_authenticate_realm = 'api'
 
-    def authenticate(self, request):
+    def get_auth_header(self, request):
         """
-        Returns a two-tuple of `User` and token if a valid signature has been
-        supplied using JWT-based authentication.  Otherwise returns `None`.
+        Returns the authentication header string split into
+        [auth_header_prefix, token] or `None` if there is no authentication header.
+
+        Raises exceptions.AuthenticationFailed() if the auth header is malformed.
         """
         auth = get_authorization_header(request).split()
         auth_header_prefix = api_settings.JWT_AUTH_HEADER_PREFIX.lower()
@@ -48,6 +50,18 @@ class JSONWebTokenAuthentication(BaseAuthentication):
             msg = ('Invalid Authorization header. Credentials string '
                    'should not contain spaces.')
             raise exceptions.AuthenticationFailed(msg)
+
+        return auth
+
+    def authenticate(self, request):
+        """
+        Returns a two-tuple of `User` and token if a valid signature has been
+        supplied using JWT-based authentication.  Otherwise returns `None`.
+        """
+        auth = self.get_auth_header(request)
+
+        if auth is None:
+            return None
 
         try:
             payload = jwt_decode_handler(auth[1])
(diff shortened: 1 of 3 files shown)


jpadilla:
@DavidMuller yeah this is what I thought would happen. If there's no sensible way around this without having users need to implement something specific to make it work, we probably shouldn't include it. Just doesn't feel right.

We're also forgetting about `JSONWebTokenAuthentication()` which also needs to decode the token.

We can definitely attempt to make it work, but only if user doesn't need to do so much work around it, otherwise I'd prefer to leave that implementation detail completely to whoever needs it.

DavidMuller:
Thanks for taking the time to read through the PR @jpadilla 

In the spirit of over-communication, I should clarify that my primary intention is *not* to add support for per-user secret keys to the package.  

I think what I really had in mind is the following: improve encapsulation of functionality in the package so that someone who wanted to implement a custom encode/decode scheme (per-user secrets, for example) could subclass relevant classes, leverage the pre-existing class methods, and then supply custom `kwargs` to the encode/decode handlers.   This would allow them to take advantage of the views + tests + configuration settings of this package instead of starting from scratch and rolling their own solution completely!

In bullet points, I would clarify my goals to be the following: 

- Encode/decode handlers that accept arbitrary `**kwargs`
- Better separation of concerns in `JSONWebTokenAuthentication`, `RefreshJSONWebTokenSerializer`, etc. such that someone could easily subclass them, leverage their pre-existing methods, and supply custom `kwargs` to the encode/decode handlers

I totally understand if those goals fall outside of the scope of this package, but wanted to share nonetheless

jpadilla:
@DavidMuller got ya. A couple of things that would then make sense for this would be:
- Add `**kwargs** instead of `user=None` to the handlers.
- Don't add user by default to the uses of the handlers. We'll just allow users that extend the package to be able to send arbitrary data to those handlers.

This is the least invasive and I'm still not sure why would someone still want to use the hooks for the handlers at that point of customizing so much, instead of using their own functions. Since they still would have to modify all the serializers and the authentication classes as well.

> This would allow them to take advantage of the views + tests + configuration settings of this package instead of starting from scratch and rolling their own solution completely!

True, but not sure to what extend, at least in the case of "per-user secrets". I think that at this point I'd need to see perhaps a simple prototype example for your use case and how this changes helped you accomplish that. Thoughts?

DavidMuller:
@jpadilla thanks again for the feedback.  

> [Add **kwargs** etc.] is the least invasive and I'm still not sure why would someone still want to use the hooks for the handlers at that point of customizing so much, instead of using their own functions. Since they still would have to modify all the serializers and the authentication classes as well.

Thinking about it a little more, I agree and understand that it might make more sense for someone to write their own wrappers around [PyJWT](https://github.com/progrium/pyjwt) instead of hacking around with this package as a 'white box.'   A case of diminishing returns for both the package maintainer (you), and someone implementing their own JWT solution.

I'll go ahead and close this pull request.

PR #68, opened 2015-01-13
Add translation utils

migonzalvar:
Use translatable error strings using ugettext function and add `locale/en_US/LC_MESSAGES/django.po` file.


migonzalvar:
I need to correct the strings on the tests too.

jpadilla:
@migonzalvar awesome, thanks again!

Issue #36, opened 2014-09-23
Custom User is not handled correctly


Nadrieril:
JSONWebTokenSerializer forces to log in with username+password credentials. It is not always compatible with custom Users. Neither is the call to get_user_model() at the module level.

jpadilla:
Closing this for lack of activity. Will reopen if there are concrete actions for a next step.

PR #70, opened 2015-01-21
user import problem for custom users issue fixed

cenkbircanoglu:
if custom model used user model import create error like this:
.....
    raise AppRegistryNotReady("Models aren't loaded yet.")
django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.

### DIFF
--- a/rest_framework_jwt/authentication.py
+++ b/rest_framework_jwt/authentication.py
@@ -1,18 +1,11 @@
 import jwt
-
 from django.utils.encoding import smart_text
 from django.utils.translation import ugettext as _
 from rest_framework import exceptions
-from rest_framework_jwt.settings import api_settings
 from rest_framework.authentication import (BaseAuthentication,
                                            get_authorization_header)
 
-try:
-    from django.contrib.auth import get_user_model
-except ImportError:  # Django < 1.5
-    from django.contrib.auth.models import User
-else:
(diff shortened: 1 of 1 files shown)


cenkbircanoglu:
while starting application it reads first import part which doesnt loaded yet. because we use custom user.
Change in the installed_apps order doesnt fix the problem. If someone know how can pass this error without change import order it will be helpfull

jpadilla:
@cenkbircanoglu was it also necessary to reorder the `from rest_framework_jwt.settings import api_settings`?

cenkbircanoglu:
no it is not necessary i'm using Pycharm and it has "optimize imports" shorcuts and i have a practice(habit) that's all beacuse of it.

jpadilla:
@cenkbircanoglu since it seems you're using a custom user model, can you check this [thread](https://github.com/GetBlimp/django-rest-framework-jwt/issues/36) out. Does this fix whatever might have been going on there?

cenkbircanoglu:
This will fix that problem:
"Well the first problem is when AUTH_USER_MODEL is changed, because there is a problem with the imports. But that is not the main problem, it is easy to fix."

But i dont understand what his problem is exactly.

jpadilla:
@cenkbircanoglu thanks!

PR #77, opened 2015-02-17
Allow setting audience and issuer

### LINKED ISSUE #79: Allow setting issuer
Allow passing in an `issuer` to `verify_signature`.

### LINKED ISSUE #81: Allow passing in an audience to verify_signature.
Allow passing in an `audience` to `verify_signature`.

cancan101:
Right now there is no way to pass in the audience to validate which causes the following to fail:
https://github.com/jpadilla/pyjwt/blob/5f9bc05803508b844d3fd9c61cd5210af06eecd8/jwt/api.py#L169

closes #79 
closes #81

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -127,6 +127,8 @@ JWT_AUTH = {
     'JWT_VERIFY_EXPIRATION': True,
     'JWT_LEEWAY': 0,
     'JWT_EXPIRATION_DELTA': datetime.timedelta(seconds=300),
+    'JWT_AUDIENCE': None,
+    'JWT_ISSUER': None,
 
     'JWT_ALLOW_REFRESH': False,
     'JWT_REFRESH_EXPIRATION_DELTA': datetime.timedelta(days=7),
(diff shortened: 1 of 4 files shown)


jpadilla:
@cancan101 this is good, thanks! Could you update any docs with this new setting? Also, might be a good idea to include a simple test since this *is* adding a new "feature".

cancan101:
How do I run tests locally? This `./runtests.py --fast` fails with:
```
E       OperationalError: no such table: auth_user
```

jpadilla:
@cancan101 that's strange, still running into issues running tests?

cancan101:
This command still has the same error: `./runtests.py --fast` but the tox commands work fine: `tox -e py27-django1.7-drf3.0.0`

jpadilla:
@cancan101 hmm weird, works fine here. I'll look into what might be going on. The only things I think is missing is the test case for this, everything else looks fine.

cancan101:
Suggestion where to put the test? https://github.com/GetBlimp/django-rest-framework-jwt/blob/master/tests/test_utils.py seems like the right place.

How do I test changing the settings?

jpadilla:
Yeah that seems right.

There's an example [here](https://github.com/GetBlimp/django-rest-framework-jwt/blob/master/tests/test_views.py#L49)..

cancan101:
Tests uploaded.

docs/index.md
```diff
@@ -182,6 +184,16 @@ This is an instance of Python's `datetime.timedelta`. This will be added to `dat
...
 
+### JWT_AUDIENCE
+This is a string that will be checked against the `aud` field of the token, if present.
+
+Default is `None`(fail if `aud` present on JWT).
+
+### JWT_ISSUER
+This is a string that will be checked against the `iss` field of the token.
+
+Default is `None`(do not check `iss` on JWT).
```
jpadilla: > (do not check `iss` on JWT)

Do you mean that if `JWT_ISSUER` is `None` and `iss` is present on the JWT it won't be checked? This bit could be a little more clearer.

cancan101: I was just trying to interpret [this code](https://github.com/jpadilla/pyjwt/blob/78742f7bee089757839921d76604b6d24d6adc32/jwt/api.py#L184) which checks the iss on the token only if `issuer` is not None:
```python
    if issuer is not None:
        if payload.get('iss') != issuer:
            raise InvalidIssuerError('Invalid issuer')
```

jpadilla:
@cancan101 thanks!

PR #80, opened 2015-02-18
Allow subclassing JSONWebTokenAuthentication

cancan101:
Right now `JSONWebTokenAuthentication` assumes that the jwt will come in the header. The JWT does not require this (see [Making a service Call](https://developer.atlassian.com/static/connect/docs/concepts/authentication.html)). The jwt may come in the querystring. The ability to send the jwt in the querystring is needed in cases where the user cannot set the header (for example the src element in HTML).

This refactoring make it easy to change the source for the JWT. For example:
```python
class JSONWebTokenAuthenticationCreateQS(JSONWebTokenAuthenticationBase):
    def get_jwt_value(self, request): 
         return request.QUERY_PARAMS.get('jwt')
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -219,6 +219,17 @@ Another common value used for tokens and Authorization headers is `Bearer`.
 
 Default is `JWT`.
 
+## Extending `JSONWebTokenAuthentication`
+
+Right now `JSONWebTokenAuthentication` assumes that the JWT will come in the header. The JWT spec does not require this (see: [Making a service Call](https://developer.atlassian.com/static/connect/docs/concepts/authentication.html)). For example, the JWT may come in the querystring. The ability to send the JWT in the querystring is needed in cases where the user cannot set the header (for example the src element in HTML).
+
+To achieve this functionality, the user might write a custom `Authentication`:
+```python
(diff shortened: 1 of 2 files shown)


cancan101:
An alternative to the way I set it up with `JSONWebTokenAuthentication` extending `JSONWebTokenAuthenticationBase` would be to use a mixin and have three classes: `JSONWebTokenAuthenticationBase`, `JSONWebTokenAuthentication`, `JSONWebTokenFromHeader`.

jpadilla:
@cancan101 yeah these seems ok. It perhaps might be a good idea to add something about this to the docs, perhaps with that `JSONWebTokenAuthenticationCreateQS ` example you gave up top.

cancan101:
Are you okay with the name of the base class: `JSONWebTokenAuthenticationBase`?

jpadilla:
@cancan101 hmmm how about `BaseJSONWebTokenAuthentication` like DRF's `BaseAuthentication`.

cancan101:
docs and rename done.

cancan101:
bump

jpadilla:
@cancan101  looks great, thanks again!

PR #88, opened 2015-03-14
Use only ASCII characters in README.rst

fantastic001:
During installation, there's an UnicodeDecodeError because in README.rst there are non-ASCII characters


jpadilla:
@fantastic001 yeap, good catch. Thanks!

PR #90, opened 2015-03-26
propagate request arg in all the doc strings

kvbik:
this change was introduced in 238e14fee97403ba509022e12950a87fe9b8e8d2
but only in the code and tests, documentation was left behind


jpadilla:
@kvbik good catch, thanks!

kvbik:
thx for the merge ;)

Issue #89, opened 2015-03-25
Add support for jti claim


avimeir:
Hi all,

I'd like to add the support for jti claim to the package. The idea basically would be to store a list of expired jti's somewhere on the server side, and allow to blacklist specific tokens.

The way I'm thinking about the implementation is:

- Use pymongo to store the blacklist (rather than a django specific mongo package, makes our solution portable to vanilla python or other frameworks)
- Generate a long enough jti claim that will have negligible probability of being produced twice before the first one expired (string of 20 random ascii_letters and digits maybe?)
- Store the jti and the entire payload in mongodb, this will help us clean up the collection when a JWT is past the expiry date (so we don't grow the collection ad infinitum)
- Add a flag in api_settings.JWT_ENABLE_BLACKLIST, if True we add the jti claim to the payload in jwt_payload_handler()
- Check against the mongodb collection everytime we decode a JWT (if JWT_ENABLE_BLACKLIST is True) and if the jti is blacklisted treat it as if the JWT is expired

I have made an initial commit to a fork of the package, with my suggestion on how to implement this, you can find it here: https://github.com/avimeir/django-rest-framework-jwt/commits/master

Happy to discuss this!

Cheers,
Avi

jpadilla:
@erichonkanen maybe you'd like to check this out

erichonkanen:
Interesting, I hadn't heard of [JWT ID](http://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4.1.7) before...  seems like this overlaps with my earlier search for a way to invalidate a token. What are your thoughts on adding mongo as a dependency to this package? The implementation looks good at first glance

jpadilla:
Mongo as a dependency is definitely a no. We should probably just include a default implementation that uses Django cache backend, like how [throttling](http://www.django-rest-framework.org/api-guide/throttling/) in DRF works.

gcollazo:
I believe Django cache is not the right solution since cache contents can be evicted or lost at any time. If I understand correctly, when you blacklist a token that token should be blacklisted forever. In that case using the DB is right choice.

I suggest creating a model `DRFJWTBlacklist` and store blacklisted tokens there. With this solution enabling a token blacklist feature will require DB access on all protected HTTP request. This should be an option and not the default behavior.

erichonkanen:
@gcollazo what are you suggesting for the default behavior?

jpadilla:
[...]

PR #91, opened 2015-03-27
Adds blacklist token feature as separate subapp

erichonkanen:
Builds off work submitted by @avimeir and the discussion at https://github.com/GetBlimp/django-rest-framework-jwt/issues/89


erichonkanen:
Tested this in a local project by logging into my app with a distant expiring JWT_EXPIRATION_DATE, issuing a POST to /api/v1/auth-token-blacklist/ with the token, and then attempting to access the api (navigated to another page with api call). A 401 Unauthorized is issued and I am logged out of my ember app as expected... Let me know if anyone has input.. seems like nice feature

erichonkanen:
Should this package include an admin.py for the default black list token implementation? Should the JWTBlackListToken model include any other fields like the actual token value?

avimeir:
@erichonkanen nice work! 

I'd add the `exp` claim to the model, so that we can later clean-up the DB by removing all expired tokens.

erichonkanen:
@avimeir thanks! should the exp be a DateTimeField or CharField? I know the format it comes in as is seconds since epoch

erichonkanen:
```
def jwt_blacklist_set_handler(payload):
    """
    Default implementation that blacklists a jwt token.
    """
    jti = payload.get('jti')
    exp = datetime.fromtimestamp(payload.get('exp'))

    return models.JWTBlackListToken.objects.create(jti=jti, expires_at=exp)
```
Made it DateTimeField and just converted

erichonkanen:
Hmm ran into issue testing, I added pytz requirement but it seems the test environment doesn't pick it up? How can I configure the test environment to use it? Should the package use the requirements.txt at installation? Also, running ./runtests.py locally shows import error for rest_framework.authenticators import OAuth2Authentication

avimeir:
I ran into the same OAuth2Authentication issue, solved it by forcing an older version of DRF:
djangorestframework==2.4.3

avimeir:
I think that it happens because OAuth2Authentication was taken out of DRF and is now a separate package: http://jpadilla.github.io/django-rest-framework-oauth/

erichonkanen:
@jpadilla ran into issue with testing, it seems like this should have popped up before but the tests/test_authentication has a call to rest_framework.authentication.OAuth2Authentication and that was moved to your 3rd party package, I adjusted the repo to use that but when running tox locally and through git I get a confusing error:

```Collecting setuptools==9.1
  Using cached setuptools-9.1-py2.py3-none-any.whl
Collecting Django==1.7.1
  Using cached Django-1.7.1-py2.py3-none-any.whl
Collecting djangorestframework-oauth==1.0.1
  Using cached djangorestframework_oauth-1.0.1-py2.py3-none-any.whl
Collecting djangorestframework==2.4.4
  Using cached djangorestframework-2.4.4.tar.gz
Collecting pytest-django==2.6.1
  Using cached pytest_django-2.6.1-py2.py3-none-any.whl
Collecting pytz==2014.10
  Using cached pytz-2014.10-py2.py3-none-any.whl
Collecting django-oauth2-provider>=0.2.4 (from djangorestframework-oauth==1.0.1)
  Using cached django-oauth2-provider-0.2.6.1.tar.gz
Collecting django-oauth-plus>=2.2.1 (from djangorestframework-oauth==1.0.1)
  Using cached django-oauth-plus-2.2.6.tar.gz
    /Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/distutils/dist.py:260: UserWarning: Unknown distribution option: 'test_requires'
      warnings.warn(msg)
Collecting oauth2>=1.5.211 (from djangorestframework-oauth==1.0.1)
  Using cached oauth2-1.5.211.tar.gz
    Traceback (most recent call last):
      File "<string>", line 20, in <module
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
    
      File "<string>", line 20, in <module
    
    ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/9r/gp7xyl_j0l3flcwp78j7mjcr0000gn/T/pip-build-2ki3xq96/oauth2
```

If you have any ideas on that or comments on the code updates (lots of incremental commits), let me know.. functionally this is all tested and working locally except for the tox build

erichonkanen:
Resolved tox error by making test oauth dependency for py27 only (didnt notice that lib doesnt support py3)...

If anyone wants to review this that'd be great, last piece is unit tests

rest_framework_jwt/models.py
```diff
@@ -1 +1,23 @@
...
+    jti = UUIDField()
+    expires = models.DateTimeField()
+    created = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        verbose_name = _('JWT Blacklist Token')
+        verbose_name_plural = _('JWT Blacklist Tokens')
+
+    def is_expired(self):
+        now = datetime.datetime.now(pytz.utc)
```
jpadilla: Shouldn't we just use `timezone.now()`?

rest_framework_jwt/compat.py
```diff
@@ -1,11 +1,35 @@
...
     from rest_framework.serializers import Serializer
 else:
-    class Serializer(rest_framework.serializers.Serializer):
+    class Serializer(serializers.Serializer):
         @property
         def object(self):
             return self.validated_data
+
+
+def get_uuid_field():
```
jpadilla: I'm not really excited about introducing an optional requirement here just for a UUIDField. I think we can just simplify this for built-in Django 1.8 UUIDField and CharField for anything less.

rest_framework_jwt/models.py
erichonkanen: ah yes, forgot about that one :) gets rid of pytz req

rest_framework_jwt/compat.py
erichonkanen: sounds good.. less dependency

erichonkanen:
hey Im running into issue with writing tests where I cant seem to load the JWTBlackListToken model into the test database and so I see:

```
OperationalError: no such table: rest_framework_jwt_jwtblacklisttoken
```

having read [this](https://docs.djangoproject.com/en/1.7/topics/testing/advanced/#using-the-django-test-runner-to-test-reusable-applications), the runtests.py file looks foreign to me and I can't figure out how to load that model? I tried copying it from rest_framework_jwt/models.py into tests/models.py, tried overriding the utiils handlers for jwt_blacklist_get/set_handler, tried adding tests/test_settings.py

any suggestions?

```
class BlacklistTokenAuthenticationTest(TestCase):
    urls = 'tests.test_authentication'

    def setUp(self):
        self.csrf_client = APIClient(enforce_csrf_checks=True)
        self.username = 'jpueblo'
        self.email = 'jpueblo@example.com'
        self.user = User.objects.create_user(self.username, self.email)
        
        api_settings.JWT_ENABLE_BLACKLIST = True
        api_settings.JWT_BLACKLIST_GET_HANDLER = test_utils.jwt_blacklist_get_handler
        api_settings.JWT_BLACKLIST_SET_HANDLER = test_utils.jwt_blacklist_set_handler

    def test_post_blacklisted_token_failing_jwt_auth(self):
        """
        Ensure POSTing over JWT auth with blacklisted token fails
        """
        payload = utils.jwt_payload_handler(self.user)
        token = utils.jwt_encode_handler(payload)

        auth = 'JWT {0}'.format(token)
        response = self.csrf_client.post(
            '/jwt/', {'example': 'example'},
            HTTP_AUTHORIZATION=auth, format='json')

        msg = 'Token is blacklisted.'

        print "RESPONSE: ",response
        self.assertEqual(response.data['detail'], msg)
        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
        self.assertEqual(response['WWW-Authenticate'], 'JWT realm="api"')
```

jpadilla:
@erichonkanen I think you'll need to include `rest_framework_jwt` to INSTALLED_APPS in `tests/conftest.py`.

erichonkanen:
@jpadilla thanks that worked... another issue is when writing test for authentication, I was debating how to blacklist the token. I ended up just importing the default implementation's model and creating a token object for the test token which effectively blacklists it:

```
class BlacklistTokenAuthenticationTest(TestCase):
    urls = 'tests.test_authentication'

    def setUp(self):
        self.csrf_client = APIClient(enforce_csrf_checks=True)
        self.username = 'jpueblo'
        self.email = 'jpueblo@example.com'
        self.user = User.objects.create_user(self.username, self.email)
        
        api_settings.JWT_ENABLE_BLACKLIST = True

    def test_post_blacklisted_token_failing_jwt_auth(self):
        """
        Ensure POSTing over JWT auth with blacklisted token fails
        """
        payload = utils.jwt_payload_handler(self.user)
        token = utils.jwt_encode_handler(payload)

        # Create blacklist token which effectively blacklists the token.
        JWTBlackListToken.objects.create(jti=payload.get('jti'),
                                         created=now(), expires=now()) 

        auth = 'JWT {0}'.format(token)
        response = self.csrf_client.post(
            '/jwt/', {'example': 'example'},
            HTTP_AUTHORIZATION=auth, format='json')

        msg = 'Token is blacklisted.'

        self.assertEqual(response.data['detail'], msg)
        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
        self.assertEqual(response['WWW-Authenticate'], 'JWT realm="api"')
```

However I also had thought of trying to just set e.g.:

```
api_settings.JWT_BLACKLIST_GET_HANDLER = True
```

but no matter if I put that line in setUp() or in the test method, it still uses the rest_framework_jwt.utils version.

erichonkanen:
@jpadilla added the rest of unit tests. if there are any other changes that need to be made let me know. I'll update docs tonight and then it should be finished

erichonkanen:
@jpadilla updated documentation, should be good to go unless you or anyone else watching has changes/suggestions.. thx

fxdgear:
I'm by no means trying to discourage this PR. 

I just want to make on statement.. If you're using JWT's as SSO. If you have another service that's got a shared secret/cert/etc... and can validate the JWT without having to make a request back to the "auth service", you're going to have to create a way to centrally store these blacklisted JWT's, and then add the overhead of EVERY request needing to ask the central auth service, if this JWT is blacklisted or not. 

For single service sites, I have no problem with this, it's a fine idea. I just want anyone who's following along to know that this won't work/scale if you're doing SOA. 

That being said I'm going to try and implement the refresh token as outlined by Auth0. https://auth0.com/docs/refresh-token

Will hopefully have an initial bare bones PR ready to go in  a couple days and hoping for some strict and constructive criticism.

erichonkanen:
@fxdgear very interesting point you bring up... This PR was result of [this](https://github.com/GetBlimp/django-rest-framework-jwt/issues/89) discussion/issue that came up and it indeed tries to implement the concept of "jti" per the json spec ([another link](https://auth0.com/blog/2015/03/10/blacklist-json-web-token-api-keys/)).

I don't have experience with SOA's personally that's something new to me... I feel like having this as an optional feature in this lib with the ability to implement the get/set on your own as overall good. I'm wondering if there are any changes though other than clear statement on the docs regarding SOA? it seems like blacklisting an auth token could be a relatively common thing for sites with stricter auth rules

fxdgear:
@erichonkanen I just think it's worth noting, for anyone who looks to implement this feature. Like I said, I don't think it's a bad idea. I am just in the situation where I need to be able to "invalidate" a JWT, but that same JWT is useable across a variety of services, and I don't want to have to make extra calls between services on every single request. 

So the route I'm gonig to take is:  
1. Submit a PR to create the ability to have "Long lived" refresh tokens
2. Change the expire time of my refresh tokens to be say 15 minutes. 
3. All consumers of my API will have to safely store the refresh tokens.
4. All consumers of my API will be able to get a new JWT without having to store the user/name password locally.

erichonkanen:
@fxdgear ok gotcha. I'll update the documentation to point that out... thanks for the input!

fxdgear:
@erichonkanen hey question I'm trying to get the unittests to run. but I'm running in to errors about
```
There is no South database module 'south.db.sqlite3' for your database. Please either choose a supported database, check for SOUTH_DATABASE_ADAPTER[S] settings, or remove South from INSTALLED_APPS.
```

South isn't in the installed apps :( Did you run into this at all?

fxdgear:
@erichonkanen I made this suggestion in my PR: https://github.com/GetBlimp/django-rest-framework-jwt/pull/94#issuecomment-88949043

maybe move the blacklist into it's own "subapp" so that the creation of the DB table doesn't happen by default, and it's only created when the `rest_framework_jwt.blacklist` is added to installed apps.

erichonkanen:
@fxdgear regarding the south error, I'm not sure what would cause that? are you running the ./runtests.py from within a virtualenv in /django-rest-framework-jwt/ ?

Regarding moving it into a subapp, I think that could be an option but id like to hear feedback from a few others first... As it is right now, the blacklist feature (and therefore the default implementation with the JWTBlacklist model) aren't installed/activated unless a user sets `JWT_ENABLE_BLACKLIST` to True (default is False).

Def open to more input

fxdgear:
I can run the tests in Tox so doesn't matter much :/ but yes I was running `./runtests.py` in the virtualenv. I ran `python setup.py develop` and `pip install -r requirements.txt` to make sure that any missing requirements were installed. 

And I'm open to suggestions too. I just think that, even if the `JWT_ENABLE_BLACKLIST` is False, the DB tables still get created.

erichonkanen:
@fxdgear hmm yeah that's weird about the south issue.. 

I see what you mean, for some reason I was thinking they weren't created unless you added "rest_framework_jwt" to installed apps but yeah django just detects the models file and installs them... I think that would be fine moving it to a subapp

erichonkanen:
@fxdgear I updated this PR to separate the blacklist token into a subapp as you advised.. seems to work nicely. I also added a note to the blacklist section of the docs but if you have more informative verbiage I can add regarding SOA let me know and I'll update it.

@jpadilla this is ready to go whenever you get a chance to run through it... thanks!

docs/index.md
```diff
@@ -116,6 +117,30 @@ Passing a token to the verification endpoint will return a 200 response and the
...
 
+## Blacklist Token
+If `JWT_ENABLE_BLACKLIST` is True, and `rest_framework_jwt.blacklist` has been added to `INSTALLED_APPS`, tokens can be made invalid (prior to expiration) by blacklisting them. More information on the concept of JTI (JWT ID) and blacklisting tokens can be read [here](http://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#jtiDef) and [here](https://auth0.com/blog/2015/03/10/blacklist-json-web-token-api-keys/).
+
+This package comes with a default implementation that stores the blacklisted tokens in the configured django database and includes an admin integration.
+
+To use this feature, add a URL like so:
+
+```python
+    url(r'^api-token-blacklist/', 'rest_framework_jwt.views.blacklist_jwt_token'),
```
fxdgear: does this doc need to be updated to point to the view as belonging to the blacklist app?

docs/index.md
```diff
@@ -116,6 +117,30 @@ Passing a token to the verification endpoint will return a 200 response and the
...
+$ curl -X POST -H "Content-Type: application/json" -d '{"token":"<EXISTING_TOKEN>"}' http://localhost:8000/api-token-blacklist/
+```
+
+If the blacklisting was successful, the response will contain the default implementation response data which is the token and a success message. Any future requests using that token will be denied.
+
+The typical use case for this feature is forcefully logging a user out due to inactivity. Many applications, especially ones with sensitive information, may implement an activity-based countdown timer and wish to instantly inactivate the user's auth token. The default implementation stores a record in the database with the unique JTI (JWT ID). However there are configurable handlers for getting and setting the blacklisted token which leaves it up to the user to decide how or where they are stored. The only requirements are that both `JWT_BLACKLIST_GET_HANDLER` and `JWT_BLACKLIST_SET_HANDLER` return a valid blacklisted token or None.
+
+**Note**
+Applications that are built with a Serivce-Oriented-Architecture (SOA) may not be able to use the blacklist token feature.
+
```
fxdgear: Would add something about significant overhead to query the auth service on every request to check if the JWT is blacklisted. (Which defeats the purpose of using JWT's anyway)

docs/index.md
```diff
@@ -136,6 +161,15 @@ JWT_AUTH = {
...
 
+    'JWT_BLACKLIST_GET_HANDLER':
+    'rest_framework_jwt.utils.jwt_blacklist_get_handler',
+
+    'JWT_BLACKLIST_SET_HANDLER':
+    'rest_framework_jwt.utils.jwt_blacklist_set_handler',
+
+    'JWT_BLACKLIST_RESPONSE_HANDLER':
+    'rest_framework_jwt.utils.jwt_blacklist_response_handler',
+
```
fxdgear: same about all these utils... should they belong in the blacklist app as well?

docs/index.md
erichonkanen: yes I will update that to make it clear

docs/index.md
erichonkanen: hm they could, although it might be easier to keep them grouped with the other utils(?) would like @jpadilla input on this

erichonkanen: I think it does make sense to separate the views and utils into the subapp as well so its completely decoupled

fxdgear: :+1:

jpadilla: Yea I guess it makes sense for these to be separate as well.

erichonkanen:
getting import error when I separated the utils/view into blacklist app:

`ImportError: Could not import 'rest_framework_jwt.blacklist.utils.jwt_blacklist_get_handler' for API setting 'JWT_BLACKLIST_GET_HANDLER'.`

erichonkanen:
Made updates, separated all blacklist related and tests passing...

docs/index.md
jpadilla: Do we really need to keep `JWT_ENABLE_BLACKLIST` around or would just adding `rest_framework_jwt.blacklist` to `INSTALLED_APPS` implicitly mean you want to enable blacklist?

erichonkanen: well, it is used in rest_framework_jwt/authentication.py and serializers.py to conditionally check for blacklisted token... I could check instead that the blacklist app is in installed_apps.. what do you think?

jpadilla: Shouldn't `BlacklistJSONWebToken` be under the `rest_framework_jwt.blacklist`? I think I'd prefer just checking if the blacklist app is installed.

erichonkanen: yes `BlacklistJSONWebToken` is under the blacklist app but the actual checks for blacklisted tokens still occur inside the authenticate method in authentication.py and inside _check_payload in serializers.VerificationSerializer.

I agree we can drop that setting but I ran into an issue where the unit tests break. I changed the check within above files to be:
```
        if 'rest_framework_jwt.blacklist' in settings.INSTALLED_APPS:
```
But in the unit tests previously I was settings `JWT_ENABLE_BLACKLIST = True` for the blacklist tests. I removed that line and get an error now saying the blacklist app is not installed (error message from check in blacklist.serializers.BlacklistJSONWebTokenSerializer)

The blacklist app is installed in tests/conftest.py, how can I get it to work with the unit test?

Previously this is what a test looked like but now removed line for `JWT_ENABLE_BLACKLIST`:
```
class BlacklistJSONWebTokenTests(TokenTestCase):

    def test_blacklist_jwt_successful_blacklist_enabled(self):
        api_settings.JWT_ENABLE_BLACKLIST = True

        client = APIClient(enforce_csrf_checks=True)

        user = User.objects.create_user(
            email='jsmith@example.com', username='jsmith', password='password')

        token = self.create_token(user)

        # Handle blacklisting the token.
        response = client.post('/auth-token-blacklist/', {'token': token},
                               format='json')

        msg = 'Token successfully blacklisted.'

        self.assertEqual(response.data['message'], msg)
```

erichonkanen: @jpadilla I just uploaded the change (tests broken) so maybe you can take a look? Im not sure why it's saying its not installed

erichonkanen:
@jpadilla resolved issues/merge conflicts, should be good to go... thx

avimeir:
@jpadilla do you have an estimate if/when this will be merged?

erichonkanen:
@fxdgear hola, I'm involved in a new side project that is looking into using microservice architecture.. I'm just learning about it etc but one of my partners suggested not using django at all meanwhile I'm promoting using rest framework etc.. I'm wondering if you could share any more info about your use case or the stack/architecture you use since you must be using django for something? (eric@epicowl.com) thanks!

jpadilla:
@avimeir @erichonkanen I'll try to review this and other pending PRs during this and next week.

erichonkanen:
@jpadilla ok awesome sounds good! did you go to pycon? watched a video from talk that mentioned jwt / rest framework etc

rest_framework_jwt/views.py
```diff
@@ -23,6 +20,7 @@ class JSONWebTokenAPIView(APIView):
     authentication_classes = ()
     parser_classes = (parsers.FormParser, parsers.JSONParser,)
     renderer_classes = (renderers.JSONRenderer,)
+    response_payload_handler = staticmethod(jwt_response_payload_handler)
```
jpadilla: Why the need for `staticmethod` here?

jpadilla:
@erichonkanen mind rebasing this again?

erichonkanen:
@jpadilla sure, do I just submit a new PR for it?

jpadilla:
@erichonkanen just fetch upstream remote, rebase, and push.

rest_framework_jwt/views.py
erichonkanen: using a static method there allows defining [custom response_payload_handler](https://github.com/erichonkanen/django-rest-framework-jwt/blob/jti/rest_framework_jwt/blacklist/views.py#L14) on a per-view basis w/o having to completely [rewrite](https://github.com/erichonkanen/django-rest-framework-jwt/blob/jti/rest_framework_jwt/views.py#L31) the post() method and not binding the method which would cause the view object to be passed into the method. Since the data being returned by that doesn't depend on the view object and lives in the utils file, using static method there seemed like good use case (ran that by #python as well)..

tldr - it didn't work w/o redefining entire post method if I didn't use static method, saves some repeat

erichonkanen:
@jpadilla ran git rebase, fixed conflicts but Im noticing this effectively made a 2nd copy of every commit now at the front.. is this expected or did I do it wrong? Ive mostly just used rebase to squash commits in the past...

jpadilla:
@erichonkanen oh snap, could you perhaps squash all of this into one commit then?

erichonkanen:
Cleaned commit history and submitted in [new PR](https://github.com/GetBlimp/django-rest-framework-jwt/pull/115)

uber1geek:
How can i implement this feature @erichonkanen ?

PR #94, opened 2015-04-02
Long running refresh tokens

fxdgear:
Taking an example of what Auth0 has done: https://auth0.com/docs/refresh-token I've tried to implement a similar process. 

This allows for a client to request refresh tokens. These refresh tokens do not expire.
They can be revoked (deleted). When a JWT has expired, it's possible to send a request
with the refresh token in the header, and get back a new JWT. This allows for the client
to not have to store username/passwords. So, if the client gets a responce about an expired token
the client can automatically make a call (behind the scenes) to delegate a new JWT using
the stored refresh token. Thus keeping the 'session' active.

**Note** This is mostly for discussion right now. But I have tested the views only (and by sideffect the serializers and authentication backend). I have not had time yet to try it on python3 though.

### DIFF
--- a/rest_framework_jwt/authentication.py
+++ b/rest_framework_jwt/authentication.py
@@ -2,12 +2,13 @@ import jwt
 from django.utils.encoding import smart_text
 from django.utils.translation import ugettext as _
 from rest_framework import exceptions
-from rest_framework.authentication import (BaseAuthentication,
-                                           get_authorization_header)
+from rest_framework.authentication import (
+    BaseAuthentication, get_authorization_header, TokenAuthentication
+)
 
 from rest_framework_jwt import utils
 from rest_framework_jwt.settings import api_settings
-
+from rest_framework_jwt.refreshtoken.models import RefreshToken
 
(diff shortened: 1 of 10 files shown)


fxdgear:
Possible suggestion: Move this long running refresh token into it's own subdir/sub-app. Like how DRF has done with `authtoken`. This would make using the long running refresh token optional, and not create a DB table if not necessary.

jpadilla:
@fxdgear how is this different or better than the already existing `RefreshJSONWebToken` that was implemented a while back?

fxdgear:
If the JWT expires in the current implementation then you have to sign in again. 

This will allow an app to be signed in "forever" without having to a) store the username/password and b) be online 100% of the time to keep the JWt active.

jpadilla:
@fxdgear haven't forgotten about this. I guess the only thing this might be need is docs and perhaps separating it into a subapp like #91.

auvipy:
@fxdgear This branch has conflicts that must be resolved

fxdgear:
closing in favor of #123

PR #95, opened 2015-04-02
Test against django1.8 and drf 3.1.0

ticosax:
Add official support for django1.8 and drf3.1


jpadilla:
Damn, I think it's time we cleaned up requirements/supported versions a bit. Thanks!

PR #99, opened 2015-04-14
add python 3.2 to tox tests

JocelynDelalande:
as support is claimed on docs (and works).


jpadilla:
@JocelynDelalande thanks!

PR #101, opened 2015-04-17
Made the email attribute optional on custom user model

JocelynDelalande:
Django itself do not not enforce requirement for an email attribute, I don't see a reason to make it mandatory here :)

### DIFF
--- a/rest_framework_jwt/utils.py
+++ b/rest_framework_jwt/utils.py
@@ -22,12 +22,15 @@ def jwt_payload_handler(user):
     except AttributeError:
         username = user.username
 
-    return {
+    payload = {
         'user_id': user.pk,
-        'email': user.email,
         'username': username,
         'exp': datetime.utcnow() + api_settings.JWT_EXPIRATION_DELTA
     }
+    if hasattr(user, 'email'):
+        payload['email'] = user.email
+
+    return payload
 
 
 def jwt_get_user_id_from_payload_handler(payload):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -34,6 +34,13 @@ class UtilsTests(TestCase):
         self.assertEqual(payload['username'], self.username)
         self.assertTrue('exp' in payload)
 
+    def test_jwt_payload_handler_no_email(self):
+        del self.user.email
+        payload = utils.jwt_payload_handler(self.user)
+
+        self.assertTrue(isinstance(payload, dict))
+        self.assertFalse('email' in payload)
+
     def test_jwt_encode(self):
         payload = utils.jwt_payload_handler(self.user)
         token = utils.jwt_encode_handler(payload)


jpadilla:
@JocelynDelalande true that, I'd prefer to actually remove that `email` by default instead of optionally adding it, but to do that I'm thinking we need to deprecate it properly, since possibly any token out there would be invalidated by this change. I'm also wondering if we still want that `username` there as well by default. I think we should go with the minimum possible which in this case would be that `user_id` which we need for actual authentication.

Thoughts?

JocelynDelalande:
Thanks for quick answer :)

> @JocelynDelalande true that, I'd prefer to actually remove that `email` by default instead of optionally adding it, but to do that I'm thinking we need to deprecate it properly, since possibly any token out there would be invalidated by this change.

Yep, another thing, if Im not wrong (but i'm quite new at JWT things) is that clients can expect email to be in the token, so better deprecate it properly ; meanwhile, I think that my patch may do the job meanwhile. Sadly, there is no way to display a deprecation warning cause we don't know if the email included is used or not.

> I'm also wondering if we still want that `username` there as well by default. I think we should go with the minimum possible which in this case would be that `user_id` which we need for actual authentication.

The minimum requirement is the username (or at least, the one defined in `USERNAME_FIELD`), not the user_id (which defaults to `user.pk` in drf-jwt code) :

> Django expects your custom User model to meet some minimum requirements. 
> Your model must have a single unique field that can be used for identification purposes. This can be a username, an email address, or any other unique attribute. 
> Your model must provide a way to address the user in a “short” and “long” form. The most common interpretation of this would be to use the user’s given name as the “short” identifier, and the user’s full name as the “long” identifier. However, there are no constraints on what these two methods return - if you want, they can return exactly the same value.

(from [django docs](https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#specifying-a-custom-user-model))

JocelynDelalande:
Thanks for quick answer :)

> @JocelynDelalande true that, I'd prefer to actually remove that `email` by default instead of optionally adding it, but to do that I'm thinking we need to deprecate it properly, since possibly any token out there would be invalidated by this change.

Yep, another thing, if Im not wrong (but i'm quite new at JWT things) is
that clients can expect email to be in the token, so better deprecate it
properly ; meanwhile, I think that my patch may do the job meanwhile.
Sadly, there is no way to display a deprecation warning cause we don't
know if the email included is used or not.

> I'm also wondering if we still want that `username` there as well by default. I think we should go with the minimum possible which in this case would be that `user_id` which we need for actual authentication.

The minimum requirement is the username (or at least, the one defined in
`USERNAME_FIELD`), not the user_id (which defaults to `user.pk` in
drf-jwt code) :

> Django expects your custom User model to meet some minimum requirements.
> Your model must have a single unique field that can be used for identification purposes. This can be a username, an email address, or any other unique attribute.
>    Your model must provide a way to address the user in a “short” and “long” form. The most common interpretation of this would be to use the user’s given name as the “short” identifier, and the user’s full name as the “long” identifier. However, there are no constraints on what these two methods return - if you want, they can return exactly the same value.

(from [django docs](https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#specifying-a-custom-user-model))

jpadilla:
@JocelynDelalande I think instead of introducing this we might be better off introducing the deprecation warnings about the payload not containing all the same fields like now in a future release. We currently only use `user_id` to actually fetch the user, but I'm inclining more to remove that as well and only keep the `username` and the `exp` in the payload. Will then fetch users with `User.objects. get_by_natural_key(username)` were appropriate.

I think a general DeprecationWarning in `jwt_payload_handler` will do.

Also, for anything else, providing a custom `jwt_payload_handler` is pretty easy.

jpadilla:
@JocelynDelalande check out https://github.com/GetBlimp/django-rest-framework-jwt/commit/816ac03629c5eca495b692f2813ae694a3c277fb

JocelynDelalande:
@jpadilla Seems pretty fine to me, thanks :). I added a comment on the commit.

PR #109, opened 2015-04-29
Verify should not refresh

### LINKED ISSUE #106: verify token returns a fresh jwt token.
Shouldn't it return the same token on success instead ?

ticosax:
fixes #106 

This pr is built on top of #108.
It should be rebased after #108 is merged (if any).
Just create the PR for review.


jpadilla:
@ticosax yeah, this looks right. thanks!

PR #110, opened 2015-04-29
Refresh from expired token

ticosax:
Allow an expired token ( `now()` > `exp` + `JWT_EXPIRATION_DELTA` ) to be refreshed.

This pr is built on top of #108.
It should be rebased after #108 is merged (if any).
Just create the PR for review.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ pytest-django==2.6
 pytest==2.5.2
 pytest-cov==1.6
 flake8==2.2.2
+freezegun==0.3.2
 
 # Optional packages
 django-oauth-plus>=2.2.1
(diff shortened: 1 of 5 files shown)


heroandtn3:
-1

This idea is useless

jpadilla:
Not sure if useless, but not sure why you'd want this. Definitely sounds insecure.

jpadilla:
@ticosax what exactly is your goal here? I would consider this an insecure practice. By allowing tokens that have already expire, you'll end up with many tokens(that have already expired) to be valid. You might also be interested in checking #94 out.

ticosax:
I believed it was the intended behaviour by reading the documentation.
I didn't plan to introduce a new flow to refresh tokens.
So I'm closing.

PR #113, opened 2015-05-05
Allow to lookup users based on a custom attribute.

ticosax:
Useful for custom User models that define another unique id field for instance.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -149,6 +149,7 @@ JWT_AUTH = {
     'JWT_REFRESH_EXPIRATION_DELTA': datetime.timedelta(days=7),
 
     'JWT_AUTH_HEADER_PREFIX': 'JWT',
+    'JWT_USER_LOOKUP_FIELD': 'pk',
 }
 ```
 This packages uses the JSON Web Token Python implementation, [PyJWT](https://github.com/jpadilla/pyjwt) and allows to modify some of it's available options.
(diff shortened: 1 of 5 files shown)


jpadilla:
@ticosax working on something that is a bit related to this on https://github.com/GetBlimp/django-rest-framework-jwt/commit/48b2c086d71d485b75f100a83179324b7feb1ece and #101.

ticosax:
Thanks I will give it a try

ticosax:
@jpadilla I tested you current dev branch.
While It technically works well, it doesn't serve my needs.
Let me explain in detail:
The field I'm relying to perform user lookup is a `UUIDField`.
Following the intentation of the patch in dev branch, I define `USERNAME__FIELD = 'uuid'` on my User model.
 It means that from now on, every django app that will consume the `User` model will use `uuid` instead of `username`. The problem is that it doesn't work quite well with all the third party libraries I'm using (`python-social-auth` for instance),  because most of the time, `username` field is assumed to be a `CharField`. It means some libraries will try to create a `User` by providing `username` that is not suitable for `UUIDField`.

To recap, while it is working fine with drf-jwt, it doesn't play well with other third party libraries.

jpadilla:
After reading this again I think there's still room for improvement here, probably by introducing an overridable function/setting that can change the current behavior in `VerificationBaseSerializer._check_user` and `BaseJSONWebTokenAuthentication. authenticate_credentials`.

PR #119, opened 2015-05-13
Some specialized serializers needs the request in context.

ticosax:


jpadilla:
@ticosax what if instead we did something like what `GenericAPIView` does providing `get_serializer()`, `get_serializer_class()`, and `get_serializer_context()`?

ticosax:
Sounds good. I'll work on it from next Monday.
I'll keep you updated.

ticosax:
@jpadilla 
I almost copy pasted the code from GenericAPIView. and provided `request` and `view` entries in context. May be you prefer I keep it empty like it is today ?

jpadilla:
@ticosax that's fine by me, thanks!

PR #121, opened 2015-05-19
Added the ability to use custom renderers for the JWT endpoints.

Jwpe:
Ran into a use-case today where I wanted to modify the format of the error response from a failed login attempt. I've already written a DRF custom renderer to neatly wrap my responses and errors, and I thought it would be cool if I could apply this renderer to the JWT endpoints as well. 

As far as I can tell it's an extremely simple change! Would definitely appreciate thoughts and feedback though.

### DIFF
--- a/rest_framework_jwt/views.py
+++ b/rest_framework_jwt/views.py
@@ -1,7 +1,5 @@
 from rest_framework.views import APIView
 from rest_framework import status
-from rest_framework import parsers
-from rest_framework import renderers
 from rest_framework.response import Response
 
 from rest_framework_jwt.settings import api_settings
@@ -21,8 +19,6 @@ class JSONWebTokenAPIView(APIView):
     throttle_classes = ()
     permission_classes = ()
     authentication_classes = ()
-    parser_classes = (parsers.FormParser, parsers.JSONParser,)
-    renderer_classes = (renderers.JSONRenderer,)
 
     def get_serializer_context(self):
         """


Jwpe:
@jpadilla does this seem like something you'd want to add?

jpadilla:
@Jwpe wouldn't just setting `DEFAULT_RENDERER_CLASSES` work? You can also do something like:

```python
urlpatterns = [
    url(r'^api-token-auth/', ObtainJSONWebToken.as_view(renderer_classes=[CustomRenderer]),
]
```

Jwpe:
@jpadilla the first suggestion wouldn't work, unfortunately, because the `render_classes` on the base view is hard-coded as `+        'rest_framework.renderers.JSONRenderer`. It could be made to work if that was changed just to use DRF's `DEFAULT_RENDERER_CLASSES` - might be preferable to a specific setting just for this.

The second way does work, but in my mind it's less elegant than permitting a global setting. It requires importing the view and the custom renderer into your URLs.

Maybe a good middle ground would just be changing the base class to have:
```python
from rest_framework.settings import api_settings as drf_api_settings
...

class JSONWebTokenAPIView(APIView):
...
renderer_classes =  drf_api_settings.DEFAULT_RENDERER_CLASSES
```

jpadilla:
@Jwpe I'm thinking `parser_classes` and `renderer_classes` are better off removed. It'll just default to whatever you set on your DRF settings. Thoughts?

Jwpe:
@jpadilla that totally works. It would be a lot simpler. If you like I can change the PR to reflect that?

jpadilla:
@Jwpe that'd be great, yeah!

Jwpe:
@jpadilla modified this and rebased to take out the irrelevant changes I made before.

jpadilla:
@Jwpe thanks!

PR #117, opened 2015-05-08
Long running refresh tokens

ticosax:
this is an attempt to resume work started in #94

@fxdgear It is mostly your work here.

PR #115, opened 2015-05-05
Added jwt blacklist subapp which enables blacklisting tokens and includes a default implementation.

erichonkanen:

PR #123, opened 2015-05-26
Long running refresh tokens

ticosax:
related to #117 and #94

But change the way we obtain a new JWT. Compare to #94, now the implementation try to be more compliant with https://auth0.com/docs/refresh-token

The delegation endpoint is a `POST`, with the following body.
    
```json
{
  "client_id":       "YOUR_CLIENT_ID",
   "grant_type":      "urn:ietf:params:oauth:grant-type:jwt-bearer",
   "refresh_token":   "YOUR_REFRESH_TOKEN",
   "api_type":        "app"
}
```

instead of being a `POST` + empty body + the refresh token passed in `Authorization` header.

/cc @fxdgear

TODO before merging:

- [ ] provide a squashed migration


rest_framework_jwt/refreshtoken/migrations/0002_auto_20150515_0948.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
fxdgear: Would probably be best to squash these migrations?

ticosax: Yes, but I would like to do it once @jpadilla considers it ready for merging. Just in case some amendments are still necessary on the Model.
I will add it in the description as a reminder.

fxdgear: :+1:

jpadilla:
This looks really good, guys. `api_type` and `grant_type` aren't used anywhere really. Wondering if that would actually introduce more confusion, thoughts?

Should we keep around our current refresh feature or should this simply replace that and be the preferred way for token refresh?

As with #123 it'd be great if we could chat and figure out if we can come up with a demo for this feature.

fxdgear:
@jpadilla so I think that both have a usecase. 

In the sense of a browser, the refresh token that exists currently is a better option. 
In the sense of a native app this long running refresh token is a better option. 

The idea is that a browser shouldn't be storing the refresh tokens (imo). And similarly a native app should NOT be storing a users password.

alvinchow86:
I think the current refresh feature would still be useful for certain applications but we should make sure to distinguish the two in the documentation

One thing that's bugging me is, should we consider an option to set expiration time on these tokens? (e.g. 14 days or something). One application I was thinking this refresh token would solve (and the existing one does not) is emulating browser sessions and keeping a user logged in (even if they haven't checked in in a while). But storing an permanent refresh token isn't that appealing.

fxdgear:
I would recommend NOT using this for browser sessions. This would require the browser to store long running refresh token. Which imo is a bad idea. 

This is more useful in the case of an iPhone app or a native app or some sort of CLI interface.

alvinchow86:
Yes that's true, XSS would be an issue. This would be indeed be very useful for mobile/native apps though. 

What do you think about an optional expiration time though?

ticosax:
@alvinchow86 Expiration can be implemented by deleting the `RefreshToken` with a celery-beat cronjob or any equivalent task scheduling system.

ticosax:
@jpadilla `api_type` and `grant_type` are not used, I agree. I blindly followed the spec https://auth0.com/docs/refresh-token .
To be honest I do not know what is better to do here, either we:
 - use those fields, but how ?
 - remove those fields.
 - keep it like that.

alarrosa14:
Will this be merged soon?

erichonkanen:
hey btw I made a demo for the [blacklist](https://github.com/GetBlimp/django-rest-framework-jwt/pull/115) token feature, maybe it could be used as the general demo for this app if one doesn't exist already? lmk if anyone wants to access it (uses ember-cli on frontend) https://jwtdemo-frontend.herokuapp.com/login

jpadilla:
I honestly think that this would make a great third party package on its own. It's more ambitious than what this package was meant to be. If @ticosax or anyone wants to pick this up go right ahead. Also living apart will help with maintenance. Thanks to everyone who worked hard on this.

ticosax:
Fair enough.
I will post here the link to the new project once it is done, for the record.

ticosax:
https://github.com/lock8/django-rest-framework-jwt-refresh-token
https://pypi.python.org/pypi/djangorestframework-jwt-refresh-token

Thanks again @fxdgear !

jpadilla:
@ticosax awesome!

fxdgear:
:+1:

Darkman:
This project is pretty much the de facto standard when using JWTs with the Django rest framework. Can I ask why this was decided to be split into another project when refresh tokens seem to be a major part of using JWTs? Especially when most articles and stackoverflow answers recommend using them? I feel that this split will cause the new package to gain far less exposure, which is so critical with security applications.

Does anyone know if DRF is thinking about natively supporting JWTs?

jammed343:
I agree with @Darkman

PR #137, opened 2015-08-04
Remove policy overrides in JSONWebTokenAPIView

resalisbury:
Allows default settings for `throttle_classes`, `permission_classes`, `authentication_classes`to be inherited from django-rest-framework's `APIView`.

### DIFF
--- a/rest_framework_jwt/views.py
+++ b/rest_framework_jwt/views.py
@@ -16,9 +16,6 @@ class JSONWebTokenAPIView(APIView):
     """
     Base API View that various JWT interactions inherit from.
     """
-    throttle_classes = ()
-    permission_classes = ()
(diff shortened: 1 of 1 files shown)


jpadilla:
@resalisbury not 100% sure about removing all of them. Setting `permission_classes` and `authentication_classes` to `()` seem sensible to the default use case. If you set your DRF default settings to require authentication, then you'd have to be authenticated beforehand to obtain a token.

Overriding `throttle_classes` and setting it to `()` on the other hand *could* be dangerous as user might have expected their default throttling settings should affect this endpoint as well.

I'd accept just removing `throttle_classes`.

resalisbury:
sounds good to me!

PR #138, opened 2015-08-04
remove throttle override in JSONWebTokenAPIView

resalisbury:
Allows default settings for `throttle_classes` be inherited from django-rest-framework's `APIView`.


jpadilla:
@resalisbury thanks

Issue #136, opened 2015-08-04
Add support for using public/private keys


orf:
The default JWT_DECODE_HANDLER/JWT_ENCODE_HANDLER blindly use the JWT_SECRET_KEY. However, and quite annoyingly, if you are using public/private keys to handle the verification then you need to pass the private key in the encode_handler, and the public one in the decode_handler.

The quickest way to fix this is to modify jwt_decode_handler like so:

```
def jwt_decode_handler(token):
    options = {
        'verify_exp': api_settings.JWT_VERIFY_EXPIRATION,
    }

    key = api_settings.JWT_SECRET_KEY
    if hasattr(key, "public_key"):
        key = key.public_key()

    return jwt.decode(
        token,
        key...
```

But obviously that's not super clean, although I think it's somewhat elegant and does fix the issue. Any thoughts?

soichih:
+1 

I also need to specify separate public / private keys to be used for RSA algorithm.

I think using additional configs (instead of JWT_SECRET_KEY) might be a bit cleaner.. 

Like..
```
'JWT_PUBLIC_KEY': open('/path/to/public.pem').read(),
'JWT_PRIVATE_KEY': open('/path/to/private.pem').read(),
```

pyjwt accepts both PEM and SSH object (like you are doing?) so either format can be passed from the config.

Then, we can do something like ...

```
def jwt_decode_handler(token):
    options = {
        'verify_exp': api_settings.JWT_VERIFY_EXPIRATION,
    }

    key = api_settings.JWT_SECRET_KEY
    if hasattr(api_settings, 'JWT_PUBLIC_KEY'):
        key = api_settings.JWT_PUBLIC_KEY

    return jwt.decode(
        token,
        key,
        api_settings.JWT_VERIFY,
        options=options,
        leeway=api_settings.JWT_LEEWAY,
        audience=api_settings.JWT_AUDIENCE,
        issuer=api_settings.JWT_ISSUER,
        algorithms=[api_settings.JWT_ALGORITHM]
    )

```

I am not sure

PR #144, opened 2015-08-11
Allow configuration of public and private key via JWT_SECRET_KEY

soichih:
Please see https://github.com/GetBlimp/django-rest-framework-jwt/issues/136

This change allows user to specify RSA key pair by doing something like

```
    'JWT_SECRET_KEY': {
        'PUBLIC_KEY': open('/home/hayashis/git/auth/config/auth.pub').read(),
        'PRIVATE_KEY': open('/home/hayashis/git/auth/config/auth.key').read(),
    },
    'JWT_ALGORITHM': 'RS256',

```

### DIFF
--- a/rest_framework_jwt/utils.py
+++ b/rest_framework_jwt/utils.py
@@ -6,7 +6,6 @@ from datetime import datetime
 from rest_framework_jwt.compat import get_username, get_username_field
 from rest_framework_jwt.settings import api_settings
 
-
 def jwt_payload_handler(user):
     username_field = get_username_field()
     username = get_username(user)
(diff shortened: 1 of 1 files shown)


jpadilla:
@soichih I think I'd prefer adding two different settings for this. Don't really like the idea that `JWT_SECRET_KEY` can be two different types. Perhaps adding `JWT_PUBLIC_KEY` and `JWT_PRIVATE_KEY`.

We'd also probably need to assert that the correct `JWT_SECRET_KEY` is used for symmetric algorithms and `JWT_PUBLIC_KEY`/`JWT_PRIVATE_KEY` are used for asymmetric algos. Thoughts?

Also, we'd definitely need some tests around this.

clintonb:
@jpadilla in lieu of keeping track of which algorithms are asymmetric, would it be acceptable to default to the public/private keys if they are set and fall back to the secret key?

```python
jwt.decode(payload, api_settings.JWT_PUBLIC_KEY or api_settings.JWT_SECRET_KEY, ...)

jwt.ENcode(payload, api_settings.JWT_PRIVATE_KEY or api_settings.JWT_SECRET_KEY, ...)
```

jpadilla:
I think #200 supersedes this.

Issue #146, opened 2015-08-17
Update PyPI package to make it DRF 3.2 compatible


outime:
The current [PyPI package](https://pypi.python.org/pypi/djangorestframework-jwt/1.6.0) doesn't work with DRF 3.2 due to the removal of the deprecated `request.DATA`. There's [a one week old commit](https://github.com/GetBlimp/django-rest-framework-jwt/commit/6d8614268d60f894dad026b5603c8a2fefc35e2f) available so it's only about pushing the update to PyPI.

jpadilla:
Released [v1.7.0](https://github.com/GetBlimp/django-rest-framework-jwt/releases/tag/1.7.0).

outime:
Gracias!

jsenecal:
For some reason I still get 
> `request.DATA` has been deprecated in favor of `request.data` since version 3.0, and has been fully removed as of version 3.2.
when using latest version from pypi. Anyone else has this issue?
the fix we implemented in https://github.com/django-json-api/django-rest-framework-json-api  adapted to your code is the folowing:
```python
def get_request_data(request):
    if hasattr(request, 'data'):
        data = request.data
    else:
        # DRF < 3.2
        data = request.DATA
    return data
```
And would prevent any references to request.DATA

jpadilla:
@jsenecal hmmm weird not sure why in your case it would actually trigger the `NotImplementedError` unless accessing `request.data` raises an exception, perhaps it's better to handle it explicitly like in your example.

@tomchristie any thoughts on what would work best for this as a compat module. There are a few other packages that will probably need this as well.

jsenecal:
@jpadilla when I edit compat.py like above the code runs as it should, no exception is raised from request.DATA.

We'll be waiting for @tomchristie I guess...

tomchristie:
The example mentioned above looks better to me. The existing compat example has a bare except, so eg malformed JSON would end up raising NotImplemented instead of responding with a parse error.

jsenecal:
Thanks @tomchristie, I submitted a PR for @jpadilla to review. (#147)

jpadilla:
Thanks all, [v1.7.1](https://github.com/GetBlimp/django-rest-framework-jwt/releases/tag/1.7.1) is out.

outime:
@jpadilla mind to push it to PyPI? :)

bharathibh:
why not to use request.body

[...]

PR #147, opened 2015-08-17
Another fix for compatibility issues with DRF 3.2

jsenecal:
The reason for this fix is explained there: https://github.com/GetBlimp/django-rest-framework-jwt/issues/146#issuecomment-131822976


jpadilla:
@jsenecal thanks!

PR #151, opened 2015-08-18
Fixes #150 "AttributeError: 'module' object has no attribute 'serializers'"

### LINKED ISSUE #150: Incompatible import of 'serializers' with DRF 3.2
Using this library with django-rest-framework 3.2, I am getting an error when importing the settings which calls on compat.py. 

```from rest_framework_jwt.settings import api_settings```

Error is ```AttributeError: 'module' object has no attribute 'serializers'``` in compat.py:
```class Serializer(rest_framework.serializers.Serializer):```...

Not sure if DRF changed something but for this to work for me, I have to add an explicit import of:

```import rest_framework.serializers```

Will submit a pull request that fixes this shortly and keep the import in only the portion of code that references the >=3.0 DRF code.

### PR DESCRIPTION
**Pull request by mikebiglan**
rest_framework does not have serializers object so needs an explicit
import. Added the import into the >=3.0 code only so does not affect
older DRF versions if this is something new.

### DIFF
--- a/rest_framework_jwt/compat.py
+++ b/rest_framework_jwt/compat.py
@@ -9,6 +9,7 @@ if StrictVersion(rest_framework.VERSION) < StrictVersion('3.0.0'):
     class PasswordField(CharField):
         widget = widgets.PasswordInput
 else:
+    import rest_framework.serializers
     class Serializer(rest_framework.serializers.Serializer):
         @property
         def object(self):
--- a/rest_framework_jwt/compat.py
+++ b/rest_framework_jwt/compat.py
@@ -9,7 +9,9 @@ if StrictVersion(rest_framework.VERSION) < StrictVersion('3.0.0'):
     class PasswordField(CharField):
         widget = widgets.PasswordInput
 else:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jpadilla** [comment]
@mikebiglan thanks! I wonder if this would work as well.

```python
from distutils.version import StrictVersion

import rest_framework
from django.forms import widgets
from rest_framework import serializers


if StrictVersion(rest_framework.VERSION) < StrictVersion('3.0.0'):
    class PasswordField(serializers.CharField):
        widget = widgets.PasswordInput
else:
    class Serializer(serializers.Serializer):
        @property
        def object(self):
            return self.validated_data

    class PasswordField(serializers.CharField):
        style = {
            'input_type': 'password'
        }
```

mikebiglan:
It's likely that would work fine and be cleaner. My only reason for putting the import mid-file like that was to prevent breaking compatibility with DRF versions older than 3.0.0. That said, it looks like the travis tests cover all the permutations of supported DRF versions, so if this is passing the tests, it would be much cleaner to move it up. I'll try that now and re-push!

PR #156, opened 2015-08-28
Feature/encode decode mixin

chrisjones-brack3t:
we're becoming homies!

## Mixin based encoder and decoder

So this is a fairly large change and it's not yet 100%. However, all current tests are passing and all Python/Django/DRF versions are passing through `tox`. I have not written tests specifically for the mixin yet, but will should you all decide this is viable. I would probably also move the mixin to a separate file.

The issue I have run into with your package is that it's not flexible when it comes to changing settings per API endpoint or serializer. My need for this package is to have the standard Django users API end points for auth & verify. I also need a set of API end points for superuser auth/very & CRUD operations. Standard users will use the default secret of `SECRET_KEY`. Superusers or rather other micro services will access the superuser end points with a separate key.

To use a different secret key for my sub-class of `JSONWebTokenAuthentication` I had to create a separate `decode_handler` so that I could specify a secret other than `settings.JWT_SECRET`, then I had to sub-class `JSONWebTokenAuthentication` & override the `authenticate` method so that I could use the new decoder. This felt needlessly complex and then I have to worry about package upgrades breaking my custom `authenticate` method.

So, my idea was to provide a mixin `JWTEncodeDecodeMixin` (name isn't set in stone and it might make sense to break it into EncodeMixin & DecodeMixin) that made it easy to either set all of the JWT properties or override the getters per API to build something up dynamically. All of the original defaults exist and work as before, with the exception of the encode and decode handlers setting. It works now because I have not yet removed the functions. I have replaced all usage of those functions throughout the codebase and updated the tests.

This mixin would make it possible to easily create your own authentication sub-classes and easily set specific settings based on your needs, or if you wanted to get crazy enough, figure out those settings based on dynamic factors by overriding the methods.

Example usage:

```python
from djangorestframework-jwt.authentication import JSONWebTokenAuthentication


class SuperUserAuthentication(JSONWebTokenAuthentication):
    jwt_secret_key = "somesupersecretkey"


class SomeCrazyAuthentication(JSONWebTokenAuthentication):
    def get_jwt_secret_key(self):
        # ... do something to dynamically figure out the key
        return key
```

I have to spend some more time tomorrow checking to see if this will be backwards incompatible. I don't think so but I'm not yet positive. I also haven't written any docs yet. Those would probably be pretty extensive. Thoughts? I want to make sure this is something you'll consider merging in before I spend more time on it or pull pieces and build something custom for my own needs.

### DIFF
--- a/rest_framework_jwt/authentication.py
+++ b/rest_framework_jwt/authentication.py
@@ -8,13 +8,13 @@ from rest_framework.authentication import (
 
 from rest_framework_jwt.compat import get_user_model
 from rest_framework_jwt.settings import api_settings
+from rest_framework_jwt.utils import JWTEncodeDecodeMixin
 
 
-jwt_decode_handler = api_settings.JWT_DECODE_HANDLER
 jwt_get_username_from_payload = api_settings.JWT_PAYLOAD_GET_USERNAME_HANDLER
 
 
-class BaseJSONWebTokenAuthentication(BaseAuthentication):
+class BaseJSONWebTokenAuthentication(JWTEncodeDecodeMixin, BaseAuthentication):
     """
     Token based authentication using the JSON Web Token standard.
     """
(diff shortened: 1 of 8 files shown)


jpadilla:
Hey @chrisjones-brack3t, I'm just catching up with this now. Definitely looks like an interesting improvement and inline with keeping things flexible enough. Backwards compatibility would definitely be important to keep in mind.

eranimo:
This would be a welcome enhancement. Any update on this?

jpadilla:
Closing this for now.

PR #159, opened 2015-09-09
This fix an error when using DRF versioning.URLPathVersioning

diegueus9:
Added *args and **kwargs to method post in JSONWebTokenAPIView to avoid error when using Versioning.


jpadilla:
@diegueus9 thanks!

PR #164, opened 2015-09-23
Added polish translations.

990px:


jpadilla:
Not merging this since it might complicate maintenance a bit.

PR #165, opened 2015-09-24
Added functionality to authenticate from querystring

sandipbgt:
User can now supply token for authentication either via header or via query string


jpadilla:
Instead of this, I'd consider adding a handler like `JWT_ENCODE_HANDLER` that was for extracting the token, defaulting to the current behavior.

PR #167, opened 2015-10-02
Adds docs for contributing

kumar303:
I wasn't sure how to begin submitting a patch so I took a guess at the steps needed. Are these correct?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -263,6 +263,23 @@ payload = jwt_payload_handler(user)
 token = jwt_encode_handler(payload)
 ```
 
+# Contributing
+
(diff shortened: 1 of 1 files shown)


jpadilla:
@kumar303 thanks for this. I think we should just probably link to whatever Django REST framework's contributing guidelines are, as we follow those closely.

http://www.django-rest-framework.org/topics/contributing/https://github.com/tomchristie/django-rest-framework/blob/master/CONTRIBUTING.md

kumar303:
Yes, a link to those docs would have been helpful to me.

PR #169, opened 2015-10-09
Move authentication into a separate method

orf:
We want to look into allowing users to login with either their username or their email in our internal app. There doesn't seem to be a nice way to support this currently, but with this change the authentication is moved to a authenticate() method, which can be overriden to do whatever.

### DIFF
--- a/rest_framework_jwt/serializers.py
+++ b/rest_framework_jwt/serializers.py
@@ -41,7 +41,7 @@ class JSONWebTokenSerializer(Serializer):
     @property
     def username_field(self):
         return get_username_field()
-        
+
     def authenticate(self, credentials):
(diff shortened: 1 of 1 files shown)


jpadilla:
While this seems to be a valid enhancement the idea for using Django's `authenticate()` was to allow you to use already known authentication backends to do things like allowing user's to authenticate with either an email or a username.

orf:
You could add a whole new authentication backend for your entire application, sure, but it seems a lot simpler and neater to handle it here. In my specific case I was going to lookup the user from their given email, then call super() with the users username and the password they gave.

jpadilla:
Closing this for now.

PR #170, opened 2015-10-15
Send user_logged_in signal when a user logs in successfully

dekkers:
The standard Django login() function sends the user_logged_in when a user logs in, with this change django-rest-framework-jwt will also send the signal. This is useful if you want to log when a user logs in. The last logged in user field is also updated by a user_logged_in receiver, so that field will also be updated with this change.


diegueus9:
This failed all the tests, and I think that is a better place emit the signal in the method authenticate of the authentication backend, like I did in PR #171

jpadilla:
Closing in favor of #171

PR #171, opened 2015-10-21
Fixes #161. Emit the signal user_logged_in when the user is authenticated.

### LINKED ISSUE #161: Emit Signals
I'd be nice if JWT auth could hook to a signal, such as 'user_logged_in'.

diegueus9:


jpadilla:
@diegueus9 thanks

jeacaveo:
@diegueus9 Excellent work.

@jpadilla Any idea as to when will this be included in the pypi package?

bluetech:
I do not think that the authenticator is the appropriate place to emit this signal.

If we take ordinary Django session authentication (e.g. Django admin), it looks like:

```
user logs in  ->  signal emitted
user access resource (page)
user access resource (page)...
user logs out
user logs in  ->  signal emitted
```

With this change, the module looks like this (unless I got it wrong):

```
user obtains token
user access resource   ->  signal emitted
user access resource   ->  signal emitted...
token expires
user obtains token
```

I would expect it to look like this:

```
user obtains token   ->  signal emitted
user access resource
user access resource
token expires
user obtains token   ->  signal emitted
```

jpadilla:
I think @bluetech brings up a good point. Anyone else care to comment? Want to put a PR together?

georgerb:
@jpadilla I think it would be a better idea to emit the signal from the validate function of the JSONWebTokenSerializer class because there is where  the log in event occurs actually, what the current PR does is trigger the logged_in signal everytime the JWT token is verified (which is equivalent to trigger the signal for every request).

jpadilla:
@georgerb want to put a PR together?

georgerb:
@jpadilla Of course, but right now the project tests fails by default, do you have a contributors document or something? Cause I faced another issue trying to install one of the requirements (oauth) on python 3.4, apparently they aren't compatible with Python 3 yet.

jpadilla:
@georgerb oh yeah, let me get those fixed first.

jpadilla:
@georgerb good to go https://travis-ci.org/GetBlimp/django-rest-framework-jwt/builds/99604454

georgerb:
Awesome, once I finish something in my job I will fix my tests and submit the PR with their proper tests. Thank you pal!

diegueus9:
There's something that I don't understand. If the signal is emitted when django calls the authenticate method in the custom authentication backend why in this case is different from the normal django way?

jpadilla:
@diegueus9 the `user_logged_in` signal is actually fired by Django on the login view not in an authentication backend. In our case we're currently(since this PR) emitting the `user_logged_in` signal every time our DRF authentication backend authenticates a request. To mimic what Django is doing we need to emit the signal once the user exchanges credentials for a token. Makes sense?

diegueus9:
If I understand correctly, the right moment to emit the signal is in 'JSONWebTokenSerializer.validate', after the code calls the 'authenticate' method from 'django.contrib.auth' because in that moment the engine will use a pair username, password instead of the current line in this PR because that code it will be called everytime a token is validated, right?

jpadilla:
@diegueus9 yup

diegueus9:
Then it's make perfect sense.

blaklites:
@jpadilla I think, the signal should be put out of both the authentication classes and serializer. The serailizer can be used at various occasions and emitting user_logged_in signals during all those cases may not be always desired. Maybe it should be left to the developers to decide when to emit this signals. 

But if we had to emit a signal in the first place, ObtainJSONWebToken view maybe the right place. Comparing to any login view we generally use in Django, where the given credentials are verified and the user is logged in --> session created for further use + signals sent, the ObtainJSONWebToken can be a place where we send the signal: the credentials are verified and the user is logged in (not really)  --> token created for further use + signals sent. Still, I think leaving the user_logged_in signal usage in case of django-rest-jwt to the developers is the best.

PR #183, opened 2016-01-03
Pass input type for PasswordField to the constructor (Fixes #182)

### LINKED ISSUE #182: Password is visible in obtain_jwt_token
The Web browsable API for the obtain_jwt_token view uses the input type text for the password field, which leaves the password visible on the screen. This is a problem, if anyone else can see your monitor (e.g. when you do a demo), because it leaks your top secret password. 

Line 22 in rest_framework_jwt/compat.py tries to prevent this, but it fails, because the attribute is overwritten, when the inherited initializer of serializers.CharField is executed. 

visible

YX4iHsvP:
Instead of setting the attribute directly, I pass it along with all other arguments to the inherited constructor. Thereby it is set there and the password is hidden again.

### DIFF
--- a/rest_framework_jwt/compat.py
+++ b/rest_framework_jwt/compat.py
@@ -18,9 +18,13 @@ else:
             return self.validated_data
 
     class PasswordField(serializers.CharField):
-        style = {
-            'input_type': 'password'
-        }
+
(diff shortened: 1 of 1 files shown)


jpadilla:
@YX4iHsvP thanks

PR #191, opened 2016-01-18
Support for string view arguments to url() is deprecated and will be removed in Django 1.10

c-bata:
Support for string view arguments to url() is deprecated and will be removed in Django 1.10.
So I update documentation.


jpadilla:
Thanks

Issue #196, opened 2016-01-29
overriding jwt_get_user_id_from_payload_handler does not work in 1.7.2


serj1975:
I previously used version 1.3.1 and override 'jwt_get_user_id_from_payload_handler'. After upgrading to 1.7.2 my code never hits.

blaklites:
That setting is not used anymore, you may override utils.jwt_get_username_from_payload_handler instead.

https://github.com/GetBlimp/django-rest-framework-jwt/blob/f07772769c0f72f971daefe1d9426a90ea099a63/rest_framework_jwt/settings.py#L22

jpadilla:
Hmmm seems like a regression was introduced [here](a3b4d44d2fc34d7752793ccbade2684707bf41da). Anyone want to put together a fix? Basically I think we need to check if a custom `JWT_PAYLOAD_GET_USER_ID_HANDLER` is used, if so we should probably keep the original behavior. Not many people seemed to have been affected by this, but still could be better.

blaklites:
@jpadilla I would like to put a fix ...

blaklites:
@jpadilla one more thing, what happens, in case, we have both JWT_PAYLOAD_GET_USER_ID_HANDLER and JWT_PAYLOAD_GET_USERNAME_HANDLER customized, we chose the custom JWT_PAYLOAD_GET_USERNAME_HANDLER then I think, right? Maybe that a very rare case, but I think we should keep a logic to handle this case in the codebase

leonsas:
What's the reason for deprecating the authentication by `user_id`? Just realized upgrading to >1.6 break my current setup.

PR #202, opened 2016-02-06
Manage custom JWT_PAYLOAD_GET_USER_ID_HANDLER when provided

blaklites:
With reference to the discussion in https://github.com/GetBlimp/django-rest-framework-jwt/issues/196, this pull request uses the custom JWT_PAYLOAD_GET_USER_ID_HANDLER when provided and when custom JWT_RESPONSE_PAYLOAD_HANDLER is missing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,32 +11,39 @@ env:
     - TOX_ENV=py27-django1.8-drf3.2
     - TOX_ENV=py27-django1.8-drf3.3
     - TOX_ENV=py27-django1.8-drf3.4
+    - TOX_ENV=py27-django1.8-drf3.5
 
     - TOX_ENV=py27-django1.9-drf3.1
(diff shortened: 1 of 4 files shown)


blaklites:
@jpadilla what do you think about this fix? Can we merge this?

jpadilla:
@blaklites I don't think this is quite right yet. We'd need to handle both cases in the authentication and serializer. If using `JWT_PAYLOAD_GET_USER_ID_HANDLER` do left, if using `JWT_PAYLOAD_GET_USERNAME_HANDLER` do right in [this](https://github.com/GetBlimp/django-rest-framework-jwt/commit/a3b4d44d2fc34d7752793ccbade2684707bf41da) diff. Makes sense?

blaklites:
Ah ok, got it....let me try and come up with the complete solution.

PR #206, opened 2016-02-12
add support for emailusernames package

dfeinzeig:
the emailusernames package for django does some horrible monkeypatching, so get_user_model().USERNAME_FIELD won't return auth.User.email.  my fix tries to import the package and if present return the email field as username_field and lookup users by their email.


dfeinzeig:
The emailusernames package was a de facto standard before custom user
models. I suspect others can benefit from this as well, particularly for
projects that have been around for a while.

PR #208, opened 2016-02-19
Fix WWW-Authenticate header when non default JWT_AUTH_HEADER_PREFIX

### LINKED ISSUE #207: JWT_AUTH_HEADER_PREFIX Changes Authorization Header but Not WWW-Authenticate
`JWT_AUTH_HEADER_PREFIX` is not being used when [constructing the `WWW-Authenticate` header](https://github.com/GetBlimp/django-rest-framework-jwt/blob/df2a01c48dc0d7b804577cb9f94863b54a32d8de/rest_framework_jwt/authentication.py#L104). Currently that value is hardcoded to `JWT`.

cancan101:
Closes #207.


jpadilla:
@cancan101 thanks!

PR #213, opened 2016-02-27
Use request.data in DRF >=  3

### LINKED ISSUE #175: JSONWebTokenAPIView fails when content is not a valid JSON
Eg.:

    curl -X POST -H "Content-Type: application/json" -d '{' http://localhost:8000/api-token-auth/

In this case `compat.get_request_data` uses request.DATA which produces an error.

    `request.DATA` has been deprecated in favor of `request.data` since version 3.0, and has been fully removed as of version 3.2.

django-rest-framework v.3+
django-rest-framework-jwt v.latest

marsam:
This close #175

### DIFF
--- a/rest_framework_jwt/compat.py
+++ b/rest_framework_jwt/compat.py
@@ -5,7 +5,12 @@ from rest_framework import serializers
 from django.forms import widgets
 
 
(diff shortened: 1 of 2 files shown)


jpadilla:
@marsam ace, thanks!

PR #215, opened 2016-03-08
Adding Django 1.8 in requirements

wdroz:
There is already Django 1.8 in README.rst, but not yet in docs.


auvipy:
fixed by https://github.com/GetBlimp/django-rest-framework-jwt/pull/224

Issue #92, opened 2015-04-01
Tokens won't refresh after expiration date despite high refresh expiration delta


MasterKale:
I've implemented `rest_framework_jwt` and have it working for authentication when accessing my API. However, I'm unable to refresh any tokens if it's after the token's `exp` value despite setting `JWT_REFRESH_EXPIRATION_DELTA` to 30 days.

For example, if I have the following settings...

    JWT_AUTH = {
        'JWT_EXPIRATION_DELTA': datetime.timedelta(seconds=900),
        'JWT_ALLOW_REFRESH': True,
        'JWT_REFRESH_EXPIRATION_DELTA': datetime.timedelta(days=30),
    }

...then shouldn't I be able to refresh the token within 30 days of the token's `orig_iat` date? Am I misreading the [settings documentation](http://getblimp.github.io/django-rest-framework-jwt/#additional-settings) for `JWT_REFRESH_EXPIRATION_DELTA`?

Out of curiosity I dug into PyJWT and found out that its `decode()` function only checks that the expiration date is later than `utcnow()+leeway`:

	if 'exp' in payload and verify_expiration:
	    utc_timestamp = timegm(datetime.utcnow().utctimetuple())

	    if payload['exp'] < (utc_timestamp - leeway):
	        raise ExpiredSignatureError('Signature has expired')

Going back into `RefreshJSONWebTokenSerializer` reveals that the token isn't getting past the initial payload check (which calls PyJWT's `decode()` function) to determine if the current time is within `orig_iat + RefreshJSONWebTokenSerializer`.

Is this just a huge misunderstanding on my part as to how JWTs work?

erichonkanen:
@MasterKale what response are you getting when you try to refresh the token? do you have any leeway set? Have you tried manually hitting your refresh endpoint (prior to token expiration) to see if it works?

MasterKale:
On the client side of things, this is the response I get when I try to refresh an expired token:

	HTTP/1.0 400 BAD REQUEST
	Allow: POST, OPTIONS
	Content-Type: application/json
	Date: Wed, 01 Apr 2015 16:55:41 GMT
	Server: WSGIServer/0.2 CPython/3.4.3
	X-Frame-Options: SAMEORIGIN

	{
	    "non_field_errors": [
	        "Signature has expired."
	    ]
	}

And here's console output from Django (I added some debug `print()`s to help troubleshoot):

	System check identified no issues (0 silenced).
[...]

PR #218, opened 2016-03-15
Ignore JWT expiration when refreshing the Token: Refresh limit should…

tleguijt:
See discussion #92.
This is a proposal for a fix

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -358,7 +358,7 @@ class RefreshJSONWebTokenTests(TokenTestCase):
 
     def test_refresh_jwt_after_jwt_expiration(self):
         """
-        Test that token can be refreshed after token expiratoin but before token refresh limit
+        Test that token can be refreshed after token expiration but before token refresh limit
         """
         client = APIClient(enforce_csrf_checks=True)
 
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -356,6 +356,23 @@ class RefreshJSONWebTokenTests(TokenTestCase):
         self.assertEqual(response.data['non_field_errors'][0],
                          'Refresh has expired.')
 
+    def test_refresh_jwt_after_jwt_expiration(self):
+        """
+        Test that token can be refreshed after token expiratoin but before token refresh limit
+        """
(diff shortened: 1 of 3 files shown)


tests/test_views.py
```diff
@@ -356,6 +356,23 @@ def test_refresh_jwt_after_refresh_expiration(self):
         self.assertEqual(response.data['non_field_errors'][0],
                          'Refresh has expired.')
 
+    def test_refresh_jwt_after_jwt_expiration(self):
+        """
+        Test that token can be refreshed after token expiratoin but before token refresh limit
```
Gaffney: typo : `expiratoin` -> `expiration`

jpadilla:
I'm still not sure about this. I keep thinking there's a possible security issue allowing this. @tleguijt where was the original discussion, can't seem to find it.

oneTimePad:
+1 about security issue. If the user attempts to refresh with an expired token that should result in a forced re-login (invalid refresh); they should be forced to resend their credentials to obtain an new token. Else the whole idea of refresh becomes pointless.

tleguijt:
@jpadilla I've linked the discussion in my first comment.

I share your doubts about security, still, I tend to think this should be the intended implementation;

> A refresh token allows the application to request Auth0 to issue a new id_token directly, without needing to reauthenticate. This works as long as the refresh token has not been revoked.
from https://auth0.com/docs/refresh-token

and

> IMPORTANT ADVICE: obtaining new tokens using the refresh_token should happen only if the id_token has expired. For example, it is a bad practice to call the endpoint to get a new token every time you do an API call. There are rate limits in Auth0 that will throttle the amount of requests that can be done using the same token from a certain IP to this endpoint.

oneTimePad:
That documentation is talking about long-time refresh tokens, a special extra token. To make it more secure a long time refresh token should be generated to allow a user to refresh and fetch a new short-time token after it has expired but checks the expiration of the long time refresh token instead

ticosax:
I'm jumping to advertise the existence of such application. https://github.com/lock8/django-rest-framework-jwt-refresh-token

PR #219, opened 2016-03-21
Update Usage docs.

mkozmin:
Hi,

I've updated docs due to this warning:

`RemovedInDjango110Warning: Support for string view arguments to url() is deprecated and will be removed in Django 1.10`


jpadilla:
@mkozmin thanks!

PR #225, opened 2016-04-02
Add missing requirements that make py.test work.

LyleScott:
I was testing a change and noticed that I could not run py.test in a fresh virtualenv.


LyleScott:
Ah, i see. I will have to follow up on this tomorrow.

angvp:
I think this PR can be closed now.

Issue #246, opened 2016-07-05
Getting error 500 when attempting to pull request


mathiasose:
My fork is at https://github.com/mathiasose/django-rest-framework-jwt/commit/a62b8f6595b232dc015190cf32ea2586341ef256

I tried opening a PR on June 10th, didn't work.
Then I forgot about it for a while until today, and it still doesn't work.

If you think this change is OK, feel free to "merge" it manually.

MattHardcastle:
Looking through the comments on open issues it's clear a few people are seeing this 500 error when creating pull requests. I know I am. I submitted a ticket with GitHub on June 21 about it but haven't heard any update after their acknowledgment.

@jpadilla do you have any idea what the issue is? It seems you've been able to accept pull requests in the past.

jpadilla:
@MattHardcastle reported this issue to GitHub. Will update when I have more info.

jpadilla:
This should be resolved.

PR #255, opened 2016-08-23
APIJWTClient login with any credential format

mathiasose:
(As mentioned in #246)

We're using emails as usernames in our project, but the `APIJWTClient.login` method assumes usernames are used.

Another option would be to use `get_username_field` [1](https://github.com/GetBlimp/django-rest-framework-jwt/blob/00f76e44da4a2b62075696cce1f583dd735ff197/rest_framework_jwt/compat.py#L22), but I think this is the better solution.
It matches what `django.test.client.Client` does [2](https://github.com/django/django/blob/9baf692a58de78dba13aa582098781675367c329/django/test/client.py#L613).

### DIFF
--- a/rest_framework_jwt/test.py
+++ b/rest_framework_jwt/test.py
@@ -4,13 +4,13 @@ from rest_framework_jwt.settings import api_settings
 
 
 class APIJWTClient(APIClient):
-    def login(self, username, password):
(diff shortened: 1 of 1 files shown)


jpadilla:
@mathiasose thanks!

PR #261, opened 2016-09-12
Add Spanish translation

xavinso:


rest_framework_jwt/locale/es/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,73 @@
...
+msgstr "Cabecera de autorización inválida. La cadena de caracteres de credenciales no debe contener espacios."
+
+#: rest_framework_jwt/serializers.py:64
+msgid "Unable to login with provided credentials."
+msgstr "No se puede iniciar sesión con las credenciales proporcionadas."
+
+#: rest_framework_jwt/serializers.py:67
+#, python-brace-format
+msgid "Must include \"{username_field}\" and \"password\"."
+msgstr "Debe incluir \"{username_field}\" y \"contraseña\"."
```
carlosasj: Actually, you should not translate `"password"` to `"contraseña"` since `"password"` refers to the field in the JSON object.

powellc: @carlosasj I don't think that's true because the password string is not in brackets, so it's just the word "password" as opposed to the username_field, which is configurable and so should be a format string.

PR #263, opened 2016-09-16
Improve documentation for expired token

guillaumevincent:
Only non-expired token can be refreshed, this correction will help

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -99,7 +99,7 @@ $ curl -H "Authorization: JWT <your_token>" http://localhost:8000/protected-url/
 ```
 
 ## Refresh Token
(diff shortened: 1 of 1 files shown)


karesh:
JWT_EXPIRATION_DELTA - 5 mins
JWT_REFRESH_EXPIRATION_DELTA - one week

I got the token and after 20 mins I tried to refresh the token I got this error.
 {
    "non_field_errors": [
        "Signature has expired."
    ]
}
**In general we should be able to refresh the token upto one week.**
What is the issue ?

guillaumevincent:
@karesh **non-expired** tokens can be "refreshed" so you should refresh your token before JWT_EXPIRATION_DELTA. So every 4 minutes for example

guillaumevincent:
I don't know why this PR is not still merged

guillaumevincent:
@jpadilla thank you

PR #264, opened 2016-09-16
Payload improved, works when user have no email

vahgar:
I tried to maintain the flow even if a User is Custom one i.e if user does not have attribute like email/username it will still work.


kodeine:
Can this be merged? i have the same issue, either we need an option to define the email field or this pr is good.

PR #268, opened 2016-10-05
Allow jtw_payload_handler to work with User models that don't have an Email field

shanx:
I'm using a custom User model that doesn't have an email field. I've made the inclusion of the email field optional in jwt_payload_handler based on whether the attr exists on the model object.


shanx:
Hey there, just a ping, can these changed be merged?

angvp:
Hey @shanx thanks for this PR, sorry it took to long to review it, seems reasonable and useful to me.

@jpadilla since this PR is old want you to have the last word before merge it, :).

PR #271, opened 2016-10-20
Update support level to DRF 3.5.

mblayman:
This branch updates the tox environment to execute against DRF 3.5.


mblayman:
@jpadilla, I think this branch helps confirm my theory that master was breaking because of some weird transient error with `cryptography`. This branch makes no changes to any code and tests *more* than master so I think it's reasonable that the broken master problem is "solved."

jpadilla:
@mblayman thank you so much for looking into this and getting it done!

PR #274, opened 2016-10-24
Fix refresh tokens

### LINKED ISSUE #249: Unable to Refresh Token, error decoding signature
I can't refresh my token. I have EXPIRATION_DELTA, REFRESH_EXPIRATION_DELTA and ALLOW_REFRESH enabled in my settings, and when I attempt to send my curl request structured like in the docs:

curl -X POST -H "Content-Type: application/json" -d '{"token":"mytoken"}' http://localhost:8000/api-token-refresh/ 

I get a return error: 'non_field_errors':["Error decoding signature"]

Am I missing something?

jorrit-wehelp:
Trying to get an new token, with an 'expired' token results in an
signature error. The token is decoded with verification of 'exp' claim.
Only the orig_iat claim should be verified in this case.

Problem is the JWT_DECODE_HANDLER does not have an option to disable exp verification.
Changing the default handler would be an backwards incompatible change... not sure how you guys would want to handle that.
So in this pull it has it's own call to jwt.encode...

Fixes #249

### DIFF
--- a/rest_framework_jwt/serializers.py
+++ b/rest_framework_jwt/serializers.py
@@ -169,3 +169,27 @@ class RefreshJSONWebTokenSerializer(VerificationBaseSerializer):
             'token': jwt_encode_handler(new_payload),
             'user': user
         }
+
+    def _check_payload(self, token):
+        # Check payload valid, but exp
+        # TODO: a jwt_decode_handler with a verify_exp option?
+        options = {
+            'verify_exp': False
+        }
+        try:
+            payload = jwt.decode(
+                token,
+                api_settings.JWT_PUBLIC_KEY or api_settings.JWT_SECRET_KEY,
+                api_settings.JWT_VERIFY,
+                options=options,
+                leeway=api_settings.JWT_LEEWAY,
+                audience=api_settings.JWT_AUDIENCE,
+                issuer=api_settings.JWT_ISSUER,
+                algorithms=[api_settings.JWT_ALGORITHM]
+            )
+        except jwt.DecodeError:
+            msg = _('Error decoding signature.')
+            raise serializers.ValidationError(msg)
+
+        return payload
+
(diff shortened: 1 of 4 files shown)


kodeine:
+1 for this

kodeine:
any idea when this will be merged?

kodeine:
@jorrit-wehelp i've tried your repo with this pull request. but i am still getting `Refresh has expired.` error.
Following is my setting.

```
'JWT_EXPIRATION_DELTA': datetime.timedelta(seconds=7200),
'JWT_REFRESH_EXPIRATION_DELTA': datetime.timedelta(days=7),
```

tvanhouten:
👍

jorrit-wehelp:
@kodeine This pull only avoids the 'Error decoding signature.' error.
With this pull request the 'Refresh has expired.' is only thrown when 'orig_iat' field (original issued at timestamp) is expired. 
But do note that the 'orig_iat' field is not updated/changed when a new token is generated during refresh...
so orig_iat field functions as a kind of maximum reuse bound for the session...
Not what many people would expect from 'refresh tokens functionality', but it's another issue that I left out of this pull request.

paw1:
Is there still a bug somewhere? I'm trying to refresh the token and still getting the error.

token : eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VybmFtZSI6ImFiY0B4eXouY29tIiwidXNlclR5cGUiOjMsInVzZXJfaWQiOjN9.vaJ1Bvmq3rJ23DmTBRmAllz_80cbGIF8JAm3SESxJl0

secret:
7ll4i+q0bhjj4yfjncixjw-0%#xbdz%c+-ys%&2*27)*a+#osf

Can be checked at [(https://jwt.io/)], it shows as verified

wlngwang:
This would solve issue #92

edelvalle:
:+1:

blueyed:
https://github.com/GetBlimp/django-rest-framework-jwt/pull/348 is meant to update this.

Do you agree?

jorrit-wehelp:
Yeah, agree. This pull is old and now has merge conflicts. Superseded by #348

PR #275, opened 2016-10-24
Allow using a cookie as a transport for the token

### LINKED ISSUE #120: Support Reading the JWT from the Requests Cookies
Right now the `Authorization` header is read but the JWT could also be stored in a cookie.

moises-silva:
A new configuration 'JWT_AUTH_COOKIE' has been added. If it set
to None (the default), everything works the same.

It can be set to a string to use as the name of the cookie that
can be used to carry the token. Note the 'Authorization' header is
still accepted as the primary method, but it falls back to the
designated cookie if present.

When a token is requested and 'JWT_AUTH_COOKIE' is set, the response
will set the given cookie with the token string.

Fix #120

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -177,6 +177,7 @@ JWT_AUTH = {
     'JWT_REFRESH_EXPIRATION_DELTA': datetime.timedelta(days=7),
 
     'JWT_AUTH_HEADER_PREFIX': 'JWT',
+    'JWT_AUTH_COOKIE': None,
 }
 ```
 This packages uses the JSON Web Token Python implementation, [PyJWT](https://github.com/jpadilla/pyjwt) and allows to modify some of it's available options.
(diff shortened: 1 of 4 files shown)


genomics-geek:
Any plans for this to be included in a release?

angvp:
@jpadilla ^ ?

orf:
I'm personally against including this, as storing a JWT in a cookie kind of defeats the point of using a JWT. It's usually an either-or situation, JWT's get around some of the limitations of cookies.

moises-silva:
@orf Heh, re-reading issue #120 I see on April last year you seemed to have a clear use case for this (similar-ish to mine). If this were a largely invasive PR I think it'd make more sense to have it as a separate package, but I don't think we want to end up as the nodejs community with 3 line packages, do we? ;)

orf:
That's a good point, I forgot about that! We worked around that issue, but I guess the use case is valid in some cases. @jpadilla what do you think?

jpadilla:
Thanks for bumping this and the reminder. After reading through #120 I think we can just merge this in.

@moises-silva thanks for putting this together. The only thing we're probably missing here are a few tests to avoid breaking support in the future.

As a quick note, anyone that wants to step up and help out with maintenance will be greatly appreciated.

moises-silva:
@jpadilla Thanks, I'll see if I can put some tests together next week and submit a new PR for those.

jpadilla:
@moises-silva sounds good, will merge this now, thanks again!

genomics-geek:
Awesome! thanks @jpadilla

jpadilla:
Just released this under [v1.10](https://github.com/GetBlimp/django-rest-framework-jwt/releases/tag/1.10.0). Thank you all!

PR #277, opened 2016-10-27
Catch AttributeError

blaklites:
No named exception was handled in the get_username_field method. This fix catches the AttributeError.


blaklites:
@jpadilla what do you think about this?

blueyed:
Thank you!

PR #237, opened 2016-04-29
serializers: fix typo

glarrain:
- "log in": adverb (action)
- "login": noun

PR #281, opened 2016-11-13
Include 'pt_BR' translation

carlosasj:
That's it. Just translated the messages.

By the way, if someone knows how to translate 3rd party packages and keep the `*.po` files in the project (not in the package folder, for versioning reason) I would be really grateful if you teach me how to do that.


angvp:
Are we accepting translations @jpadilla ? I think having i18n for messages of django-rest-framework-jwt is a little bit excessive but would like to hear other opinions, do not want to waste the work done by @carlosasj

jpadilla:
@angvp at some point I thought it was a good idea, but then PRs like #237 are a bit hard to reason about. I would love to remove them in a next mayor bump.

powellc:
@jpadilla why would you remove translations? We're using this project, but need the error messages returned in Polish and are facing a position of vendoring into into our own project or publicly forking it simply to add a Polish LC_MESSAGES file.

PR #285, opened 2016-11-15
Use get_username_field to determine username field on payload

### LINKED ISSUE #284: The default `jwt_get_username_from_payload_handler()` might fail with other username field
When I updated my payload by editing the `jwt_payload_handler()` I all of a sudden got `Payload invalid` errors. This was due to the fact that I removed the `username` from the payload altogether, since in my app `email` is the username field.

I therefore believe the default `jwt_get_username_from_payload_handler()` should also use the `get_username_field()` function.

jgadelange:
Fixes #284.

I think this is a more sensible default for the `jwt_get_username_from_payload_handler()`

### DIFF
--- a/rest_framework_jwt/utils.py
+++ b/rest_framework_jwt/utils.py
@@ -62,7 +62,8 @@ def jwt_get_username_from_payload_handler(payload):
     """
     Override this function if username is formatted differently in payload
     """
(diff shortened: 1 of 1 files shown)


blueyed:
Please add a test for it.

PR #287, opened 2016-11-21
Add request parameter to jwt_decode_handler (fixes #243)

### LINKED ISSUE #243: Pass request to jwt_decode_handler in authenticate
I would like to log my JWT token use in my custom jwt_decode_handler. Having the request object available  would allow me to get at more interesting logging information. Is this a feature you would be open to adding?

MattHardcastle:
Currently only the `jwt_token` is passed to the `jwt_decode_handler`. This pull request adds the `request` parameter if `jwt_decode_handler` throws a `TypeError` while being called with only one parameter (which happens if `jwt_decode_handler` takes more than one parameter).

### DIFF
--- a/rest_framework_jwt/authentication.py
+++ b/rest_framework_jwt/authentication.py
@@ -1,3 +1,5 @@
+import inspect
+
 import jwt
 
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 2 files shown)


blueyed:
I like the idea of passing in `request`, but there might be a better way to support both function signatures?!

Also: needs to be rebased.

MattHardcastle:
Thanks for considering my pull request. I've rebased it with the latest master. I also changed the signature check method from catching the TypeError exception to actually inspecting the code object to see how many args it takes.

I initially wrote it to support both signature because I'm sensitive to the fact that it's a breaking change. But it's probably best to pass requests as a keyword argument and take the breaking change.

MattHardcastle:
I may reopen this once I get legal ok from my employer.

PR #313, opened 2017-03-02
Add ability to specify time delta and value

michaelaelise:
It would be nice to be able to create tokens and specify the expiration at runtime.

I am attempting to bridge the gap between manually creating tokens and the one time set expiration.

This attempt allows two extra parameters to be passed to obtain jwt.
 tdtype is the python timedelta keyword
 tdvalue is the timedelta value

With my changes the following syntax may be used to create a token and specify the expiration.

`curl -X POST -d "username=admin&password=default&tdtype=seconds&tdvalue=800" http://172.27.72.29:8001/api-token-auth/`

Works for 'seconds', 'minutes', 'hours', 'days', and 'weeks'.

Possible improvements:
  Only allow superusers to specify the timedelta, everyone else gets the default. This would be easy to verify in the serializer.

### DIFF
--- a/rest_framework_jwt/serializers.py
+++ b/rest_framework_jwt/serializers.py
@@ -32,15 +32,17 @@ class JSONWebTokenSerializer(Serializer):
         Dynamically add the USERNAME_FIELD to self.fields.
         """
         super(JSONWebTokenSerializer, self).__init__(*args, **kwargs)
-
         self.fields[self.username_field] = serializers.CharField()
         self.fields['password'] = PasswordField(write_only=True)
+        self.fields['tdtype'] = serializers.CharField(required=False)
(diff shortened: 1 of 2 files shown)


michaelaelise:
Pardon my newbieism but the only test that is failing contains...
`>       self.assertEqual(serializer.errors, expected_error)
E       AssertionError: {u'non_field_errors': [u'Unable to login with provided credentials.']} != {'non_field_errors': ['Unable to log in with provided credentials.']}
`
unicode vs non unicode

I did not change that test.

angvp:
Great and quick job @michaelaelise I have some stuff to review yet, I did a comment that you should start looking at it, I think there are missing docs as well, I will review it tonight and add more comments here :).

PR #323, opened 2017-04-06
Don't require the 'token' key to override jwt_response_payload_handler

brianrower:
If you have JWT_AUTH_COOKIE enabled and you cannot override jwt_response_payload_handler with your own implementation that does not include the "token" key. This small change uses the token as you already have it rather than the trying to retrieve it from the response data


jpadilla:
@brianrower this looks good to me, thanks! 🎉 🎆

Issue #258, opened 2016-09-04
Some way to use this on 2 servers


luisiturrios:
I try make a SSO on my company like this example

**JWT Server:** https://auth.domain.com/
**API:** https://api.domain.com/

PR #335, opened 2017-05-19
Allow token-cookie to be cross-domain

j-paulino:
Addresses #258. Example use case:

 Support for Api servers on one sub domain (api.example.com) and web servers on a another (app.example.com).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -179,6 +179,7 @@ JWT_AUTH = {
 
     'JWT_AUTH_HEADER_PREFIX': 'JWT',
     'JWT_AUTH_COOKIE': None,
+    'JWT_AUTH_COOKIE_DOMAIN': None,
 
 }
 ```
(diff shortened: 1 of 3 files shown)


MiguelSR:
This would be so useful... will it be accepted? Thanks.

yotammanor:
Upvoting!

butlerdav:
I agree, this will be very helpful. Currently I am subclassing JSONWebTokenAPIView just so I can add the domain kwarg to set_cookie.

jourdanrodrigues:
I totally upvote this improvement (and thanks @butlerdav for the workaround), but there's already `JWT_AUDIENCE` in the `api_settings`. Do you think it could be reused for this purpose or are they different things?

butlerdav:
@jourdanrodrigues, they are different things.`JWT_AUDIENCE` is about setting/validating the value of `aud` claim (i.e. part of the JWT spec, see https://tools.ietf.org/html/rfc7519#section-4.1.3).

KrYpTeD974:
+1 for this

PR #341, opened 2017-06-12
The verify argument is deprecated in jwt.decode

blueyed:
This uses the "verify_signature" option instead.

### DIFF
--- a/rest_framework_jwt/utils.py
+++ b/rest_framework_jwt/utils.py
@@ -98,15 +98,16 @@ def jwt_encode_handler(payload):
 
 def jwt_decode_handler(token):
     options = {
+        'verify_signature': api_settings.JWT_VERIFY,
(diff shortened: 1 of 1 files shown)


angvp:
This PR is breaking the tests, would you mind to fix it?

blueyed:
Sure.

But it looks like `verify` is not actually deprecated for `PyJWT.decode`, but only `PyJWS.decode`.
Let's wait on the outcome of https://github.com/jpadilla/pyjwt/pull/271.

jpadilla:
@blueyed can you still check this out, if not I can take over. Looking to release a new version today.

blueyed:
With https://github.com/jpadilla/pyjwt/pull/271 there are no more deprecation warning here, so this can be closed.

jpadilla:
Ace, thanks!

PR #344, opened 2017-06-22
Optimize jwt_decode_handler

blueyed:
1. There is no need to decode the payload without verification first, in
   case `api_settings.JWT_PUBLIC_KEY` is used.
2. It also skips decoding in case
   `api_settings.JWT_GET_USER_SECRET_KEY` is not used.
   This might be a problem in case somebody is monkey-patching
   `jwt_get_secret_key`, but that should not be considered stable/public
   API, should it?

I had this still from when working on https://github.com/GetBlimp/django-rest-framework-jwt/pull/341.
Might be better to keep this for after the currently planned release in any way.


blueyed:
@jpadilla 
Anything to be done here?

(the build failure seems unrelated - I've only just rebased it; should be fixed with https://github.com/GetBlimp/django-rest-framework-jwt/pull/369)

fengsi:
Oh just saw this PR.  This does not seem to handle the encode part.  I also had a PR, taking care of both:
https://github.com/GetBlimp/django-rest-framework-jwt/pull/416

PR #348, opened 2017-07-17
Don't verify token expiration on token refresh.

edelvalle:


blueyed:
Makes sense.

Can you add a test for this, please?

blueyed:
Can we use https://github.com/GetBlimp/django-rest-framework-jwt/pull/274/files#diff-a4f5c4d64d4811b15e1405573c00f3f4 for the test here?

blueyed:
I've not looked into the details much, but https://github.com/GetBlimp/django-rest-framework-jwt/issues/92#issuecomment-327746574 seems to discuss this well.

edelvalle:
@blueyed Well, it took me a while 😆 
But here are the tests 🚀

Emnalyeriar:
Any news on this? Will this be merged?

edelvalle:
I wonder the same... lol

blueyed:
Thanks for the test / completing it - but I cannot merge it myself.
You will have to patiently wait for some maintainer to review and do it.

edelvalle:
@jpadilla Jose, can you review this PR please? Gracias!

pmferreira3:
Can you please merge and release this PR?

Alex3917:
> When a token is expired and you try to refresh it inside of the refresh interval. It does not refresh, because the refresh endpoints checks if the token is not expired

AFAIK this is the intended behavior. I could be wrong, but my understanding is that users are supposed to be required to re-authenticate every so often, so that if someone else has access to their token they can't just keep using it indefinitely without doing something like changing the user's password that would lead to them being detected.

Koomook:
@Alex3917 I'm bit confused. Then why refresh mechanism exists? I totally agree with @edelvalle.

Alex3917:
> Then why refresh mechanism exists?

Think about the way your bank's website works. Once you log in, you're probably allowed to keep using the site for up to 24 hours without logging in again, but only if you're continuously using the site and requesting a new page at least once every 15 minutes.

In this scenario the access token life (`JWT_EXPIRATION_DELTA`) would be 15 minutes, and the refresh token life (`JWT_REFRESH_EXPIRATION_DELTA`) would be 24 hours.

And that's the way it's supposed to be; your bank doesn't want you to be able to keep using your current session if you've already gotten logged out for inactivity, at least not unless you enter your username and password again. Allowing people to keep requesting a new access token after it's already expired would break that behavior.

Are there use cases where just making the `JWT_EXPIRATION_DELTA` longer isn't a good solution?

pmferreira3:
Image we have the following use case of a generic API and as you mentioned we have configured:
```
JWT_EXPIRATION_DELTA= 15min 
JWT_REFRESH_EXPIRATION_DELTA = 24h
```
1. I never accessed the API before, so start clean;
2. First login so I receive the token;
3. For the next 15 min I invoked several endpoints the API passing the token;
4. Now, the token is no longer valid. 

The question is: How can I refresh the token?
I am within the 24h period but over than 15 min since the login:
- Should I be able to invoke the refresh-token endpoint and get the new one?
- Should I login again receiving a new token?
- Should the behavior of the client be refresh the token e.g. every 10 min?

Alex3917:
@pmferreira3 

The way we do it is every time we make a request from the front end to the API, there is a request interceptor that checks how much time is left until the token expires. If there is less than a certain amount of time remaining, the request interceptor triggers a function to refresh the token and then store the new token in local storage, and then continues with the original request but with the new token.

So for example if the access token lasts 15 minutes, you could refresh the token before any request where there are less than 12 minutes remaining. The reason for this is that rendering any one page could easily be 10 or 15 network requests, and it would a needless performance hit to refresh the token before each network request. But this logic still allows the token to be refreshed on every page view, unless those page views are basically right after one another.

jayvdb:
The commits which came from https://github.com/GetBlimp/django-rest-framework-jwt/pull/274 should have @jorrit-wehelp as the author in the commit metadata.

PR #359, opened 2017-08-29
test

jacoor:


jacoor:
forget it.

blueyed:
@jacoor 
Looks like you have accidentally pushed this here instead of your private repo?

In case you are still using it yourself, you might actually want to bring it upstream after all?!

jacoor:
Well, owner of the library decided they don't want to use permanent tokens. Instead, we created our own lib which extends functionality of this library using it as dependency. 
The PR here is a test PR that's why it was closed.
If you are interested in our lib: https://github.com/ArabellaTech/drf-jwt-devices

PR #360, opened 2017-08-29
Optimization of DB queries when decoding token with per user secret.

jacoor:
When using token secret on user, DB is called twice to decode token:
* once by id to get secret
* second time by natural key in "authenticate_credentials"
Because of two different keys used to call DB this database query was hard to cache.
The code was changed to use "get_by_natural_key" in both cases.

Other improvements:
* added py36 in travis - this works
* set travis to use "precise" as default distribution, otherwise tests fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ python: '3.5'
 
 language: python
 
+dist: precise
+
 cache: pip
 sudo: false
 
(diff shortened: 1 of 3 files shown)


jacoor:
Hello @jpadilla,
you probably remember me from creating https://github.com/GetBlimp/django-rest-framework-jwt/pull/310.
Now I added a little optimisation we could use on our projects soon. Could you release this in the near future, pretty please?

Thank you!

PR #363, opened 2017-09-08
support white space in password

danieldai:


blueyed:
This changes existing behavior (trimming whitespace seems to be the default for CharFields?!), which is good.

I am not sure there should be a setting for this though.

Anyway, please rebase it for now (tests are fixed on master).

Issue #356, opened 2017-08-09
Is it possible to return a new token from the refresh token endpoint?


aronysidoro:
My co-worker had raised this concern, that the token after refresh should have a new value.  Is this possible? I did try looking at the source code, and some `pyjwt` source code, but I wasn't sure.

Thank you

PR #366, opened 2017-09-08
Add setting for extending the orig_iat for token-refresh and also causes a new token value to be generated when refreshing a token

aronysidoro:
I'm submitting this PR, so that refresh token values are different every time when refreshing a token.

I had originally submitted issue #356 

I did some testing, and in fact the original token value is returned when doing a token refresh, as long as the the "orig_iat" is less than the expiration delta, where the default is 7 days.

There may be a use case, and I myself would like to have this feature, where the token value is different every time a token is refreshed.  

I added tests and documentation.  If this PR is conceptually acceptable, but maybe some changes are needed and so on, please let me know, and I'd be happy to make some changes.  I'd just like to see the overall feature land, so we can use it.

Thank you


blueyed:
Please rebase this since now the tests should be fixed again.

aronysidoro:
@blueyed thank you for the bump.  The tests passed. What do you all think about my PR?  

I hope that I addressed the feature correctly, and it's a feature that you all want.  Another person on my team had noticed that it's the same token value if the orig_iat is the same, so I wanted to make this PR.  

Thanks

blueyed:
I do not like new settings personally..

Is it related to https://github.com/GetBlimp/django-rest-framework-jwt/pull/348 even after all?

aronysidoro:
So maybe the issue is that `orig_iat` is in days, and so the same token value will persist for a minimum of 1 day, depending on how you set the setting.

What about allowing for a different time interval for `orig_iat` other than days?

crodriguez1a:
@blueyed @aronysidoro I think this PR covers two distinct issues that both deserve a second look. One was sort of covered [here](https://github.com/GetBlimp/django-rest-framework-jwt/issues/92#issuecomment-327746574)(not sure if documentation will be updated), but the other I'm still not clear on, and it seems that this PR would address it.  Should the expectation be that content of the token upon refresh would not be re-computed?

PR #370, opened 2017-09-15
Return decoded payload in `authenticate`

blueyed:
This allows for accessing it on `request.auth` easily then.

TODO:
 - [x] test (none is failing?!)


blueyed:
Should I go forward with adding a test here?

blueyed:
@jpadilla 
Added a test, please review.

Should this be documented?  Where (in general, and for the changelog)?

jpadilla:
@blueyed I failed earlier on keeping an updated changelog so I've been just doing so on each release https://github.com/GetBlimp/django-rest-framework-jwt/releases

PR #374, opened 2017-09-16
[RDY] Added support for Custom Exception Handler

ideallical:
This PR makes it possible to enable DRF's custom exception handler, by default the old behaviour is kept for backwards compatibility. Typo has been fixed.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -179,6 +179,7 @@ JWT_AUTH = {
 
     'JWT_AUTH_HEADER_PREFIX': 'JWT',
     'JWT_AUTH_COOKIE': None,
+    'JWT_USE_CUSTOM_EXCEPTION_HANDLER': False
 
 }
 ```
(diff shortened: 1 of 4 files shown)


blueyed:
Please add tests for it.

ideallical:
Should be ready for merge now, @blueyed

blueyed:
Thanks!
Can you rebase it please, so we can see how https://github.com/GetBlimp/django-rest-framework-jwt/pull/382 has an impact?

ideallical:
the branche has been rebased @blueyed

blueyed:
Maybe re-opening helps for codecov.

Otherwise I've thought that I've left a comment for this to be the default, but have not?!

blueyed:
See also #386.

ideallical:
So, coverage report says coverage went up. Time to merge? @blueyed

blueyed:
@ideallical 
Thanks, but I think this should be just the default, without a setting for this.

blueyed:
@jpadilla ?

draskomikic:
Hi guys, I am also interested in this feature, when do you expect it will be merged ?

ilstreltsov:
@blueyed 
Making this behavior (raise exception) as default will break backward compatibility, right? So i suggest  just to merge PR as it is. Up!

PR #378, opened 2017-09-22
Fix token in cookie unable to refresh

C0deZLee:


blueyed:
Needs a test.

btw: you can run `tox -e py36-flake8` locally.. ;)

the1900:
Thanks @C0deZLee You give me a big inspiration.

PR #384, opened 2017-09-28
jwt_get_secret_key check if user exists

kpourdeilami:
When a user with a deleted account tries to access the website with a JWT that is still valid, it causes an error 500. This checks if the user with the given pk exists, and if a user not, it will return the `api_settings.JWT_SECRET_KEY` instead of the deleted user's secret key

### DIFF
--- a/rest_framework_jwt/utils.py
+++ b/rest_framework_jwt/utils.py
@@ -23,9 +23,12 @@ def jwt_get_secret_key(payload=None):
     """
     if api_settings.JWT_GET_USER_SECRET_KEY:
         User = get_user_model()  # noqa: N806
-        user = User.objects.get(pk=payload.get('user_id'))
-        key = str(api_settings.JWT_GET_USER_SECRET_KEY(user))
-        return key
+        try:
+            user = User.objects.get(pk=payload.get('user_id'))
+            key = str(api_settings.JWT_GET_USER_SECRET_KEY(user))
+            return key
+        except User.DoesNotExist:
+            pass
     return api_settings.JWT_SECRET_KEY
 
 


rest_framework_jwt/utils.py
```diff
@@ -23,9 +23,12 @@ def jwt_get_secret_key(payload=None):
...
         User = get_user_model()  # noqa: N806
-        user = User.objects.get(pk=payload.get('user_id'))
-        key = str(api_settings.JWT_GET_USER_SECRET_KEY(user))
-        return key
+        try:
+            user = User.objects.get(pk=payload.get('user_id'))
+            key = str(api_settings.JWT_GET_USER_SECRET_KEY(user))
+            return key
+        except User.DoesNotExist:
+            pass
```
lahim: `jwt.InvalidTokenError()` exception should be raised, for sure there is not place for `pass` statement. Please take a look into the `BaseJSONWebTokenAuthentication` class where `InvalidTokenError` exception is handled.

rest_framework_jwt/utils.py
```diff
@@ -23,9 +23,12 @@ def jwt_get_secret_key(payload=None):
     """
     if api_settings.JWT_GET_USER_SECRET_KEY:
         User = get_user_model()  # noqa: N806
-        user = User.objects.get(pk=payload.get('user_id'))
-        key = str(api_settings.JWT_GET_USER_SECRET_KEY(user))
-        return key
+        try:
+            user = User.objects.get(pk=payload.get('user_id'))
+            key = str(api_settings.JWT_GET_USER_SECRET_KEY(user))
```
lahim: This code should be outside of this `try ... except` block. So it should look like below:

```python
try:
    user = User.objects.get(pk=payload.get('user_id'))
except User.DoesNotExist:
    raise jwt.InvalidTokenError()
else:
    return str(api_settings.JWT_GET_USER_SECRET_KEY(user))
```

detectedstealth:
Any ETA on when this will be merged? I ran into the same issue with a deleted user.

Issue #387, opened 2017-10-09
Provide `request` to contrib.auth.authenticate in serializers


gtnx:
In Django 1.11, `django.contrib.auth.authenticate` takes the current `request` as an argument. `JSONWebTokenSerializer` should support that.

jksimoniii:
Addressed https://github.com/GetBlimp/django-rest-framework-jwt/pull/412

PR #388, opened 2017-10-09
Provide `request` to contrib.auth.authenticate in JSONWebTokenSerializer

gtnx:
refs #387


rest_framework_jwt/serializers.py
```diff
@@ -3,6 +3,7 @@
 from calendar import timegm
 from datetime import datetime, timedelta
 
+from django import __version__
```
blueyed: Better just import `django` and then use `if django.VERSION >= (1, 11)`

Issue #367, opened 2017-09-13
could i add other information to payload?


vparenti:
I would like add more information in payload json, for example 'role' and 'company'.
Is there a way?
Thanks,
V

mauzepeda:
+1

Alex3917:
You just make a custom payload handler, e.g.:

```
from rest_framework_jwt.utils import jwt_payload_handler

def jwt_custom_payload_handler(user):
    payload = jwt_payload_handler(user)
    payload['role'] = 'xyz'
    return payload
```

Then in your settings you just need:

```
JWT_AUTH['JWT_PAYLOAD_HANDLER'] = 'path.to.your.jwt_custom_payload_handler',
```

blueyed:
Could we add this to the docs then?
@vparenti / @mauzepeda a PR would be welcome.

mauzepeda:
@blueyed I've added @Alex3917 example to the docs and made a Pull Request.

PR #389, opened 2017-10-09
Added JWT_PAYLOAD_HANDLER usage example to the docs

mauzepeda:
Added JWT_PAYLOAD_HANDLER usage example to the docs. https://github.com/GetBlimp/django-rest-framework-jwt/issues/367#issuecomment-334878640

@blueyed


blueyed:
Reminded me of https://github.com/GetBlimp/django-rest-framework-jwt/pull/370.

blueyed:
Maybe we should have an explit test to cover this use case?

Issue #303, opened 2017-01-29
For inactive user,  `JSONWebTokenSerializer.validate()` raises `Unable to login with provided credentials.` 


ibrahim12:
>  For inactive user,  `JSONWebTokenSerializer.validate()` raises `Unable to login with provided credentials.` 

Django Version 1.10.5

Django Model Backend's authenticate method checks as follows, 
 

    if user.check_password(password) and self.user_can_authenticate(user):
        return user

And user_can_authenticate method is 

     def user_can_authenticate(self, user):
            """
            Reject users with is_active=False. Custom user models that don't have
            that attribute are allowed.
            """
           is_active = getattr(user, 'is_active', None)
           return is_active or is_active is None

So, If user is inactive then, in `JSONWebTokenSerializers.validate` method returns wrong response as it don't goes to that block of code as user is None.

      if user: <- This is None, for inactive user.
            if not user.is_active:
                   msg = _('User account is disabled.')
                   raise serializers.ValidationError(msg)

             payload = jwt_payload_handler(user)

             return {
                   'token': jwt_encode_handler(payload),
                   'user': user
             }
       else:
              msg = _('Unable to login with provided credentials.')
              raise serializers.ValidationError(msg)

RubenSchmidt:
Is this not related to the Django 1.10 change for the ModelBackend? Inactive users are not allowed to authenticate by default any more. You have to use AllowAllUsersModelBackend (https://docs.djangoproject.com/en/1.10/ref/contrib/auth/#django.contrib.auth.backends.AllowAllUsersModelBackend) To allow inactive users to authenticate.

igsm:
I face the same issue. I still do NOT want to authenticate inactive user, however, I'd like to differentiate the error response. It always throws 'wrong credentials' error. How can I solve this problem?

mehtamanan:
this would help me a lot too

mrtysn:
I am also looking for a workaround for this problem. Is overriding the VerifyJSONWebTokenSerializer a good idea?

PR #391, opened 2017-10-25
Fix is_active error message

iliasgal:
Fix for issue #303 

If user is disabled (is_active=False), the authenticate function returns None, so the user variable in: 

`  user = authenticate(**credentials) 
`

 is always None.

Therefore, the if statement below is always False and the commands are never executed.
```
if not user.is_active:
    msg = _('User account is disabled.')
    raise serializers.ValidationError(msg)
```

My suggestion is to move the check for 
`
if not user.is_active:`
under the else statement. 

We get the user with a query based on the given username. We also add an exception if ObjectDoesNotExist. 

If user exists, then we check if it is active or not.
If the ObjectDoesNotExist is thrown, then it means that the login credentials were not correct.

```
 try:
    user = User.objects.get(**{self.username_field: attrs.get(self.username_field)})
    if not user.is_active:
        msg = _('User account is disabled.')
        raise serializers.ValidationError(msg)
    except ObjectDoesNotExist:
        msg = _('Unable to log in with provided credentials.')
        raise serializers.ValidationError(msg)
```

### DIFF
--- a/rest_framework_jwt/serializers.py
+++ b/rest_framework_jwt/serializers.py
@@ -5,6 +5,7 @@ from datetime import datetime, timedelta
 
 from django.contrib.auth import authenticate, get_user_model
 from django.utils.translation import ugettext as _
+from django.core.exceptions import ObjectDoesNotExist
 from rest_framework import serializers
 from .compat import Serializer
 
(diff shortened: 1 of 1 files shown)


mehtamanan:
Waiting for the merge :)

mehtamanan:
And just FYI, this code will not have expected behaviour in the case where someone sends a username of an inactive user but an incorrect password. 

According to the commit, the response will be 'User account is disabled.' but it should be 'Unable to log in with provided credentials.' as the password in itself is incorrect.

PR #401, opened 2017-12-06
Use validated_data to access validated data

xrmx:
Instead of object which is not documented in DRF and for some
reason does not work here.


xrmx:
Friendly ping :)

xrmx:
@blueyed sorry to bother, any chance you can merge this please?

PR #409, opened 2017-12-15
Fixing authentication

kakulukia:
not sending a request object is deprecated since Django 1.11 and this also broke Django Axes .. since the request is already in the context, thats an easy fix

### DIFF
--- a/rest_framework_jwt/serializers.py
+++ b/rest_framework_jwt/serializers.py
@@ -47,7 +47,7 @@ class JSONWebTokenSerializer(Serializer):
         }
 
         if all(credentials.values()):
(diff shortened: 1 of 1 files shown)


kakulukia:
Well, the serializer tests dont fill the request context in the same way as the JSONWebTokenAPIView is initializing its serializer .. BUT not sending a request object in the authenticate call is deprecated since Django 1.11. 
Sadly i dont know how to fake the request that easily otherwise i would fix the failing serializer test. 
Please somebody help me out here!?

jksimoniii:
Addressed https://github.com/GetBlimp/django-rest-framework-jwt/pull/412

Issue #410, opened 2017-12-19
user_logged_in signal


atmishra:
I am looking for a signal similar to user_logged_in, which should be emitted when a user is authenticated successfully. What would be the best way to implement it. Currently I am overriding JSONWebTokenSerializer validate() method to emit signal. But it's leading to code redundancy.

PR #413, opened 2017-12-28
refactor views to provide serializer_valid and serializer_invalid as …

ralokt:
…hooks for subclasses.

I'm using this app for a project of my own, and wanted to set `user.last_login`.

I felt like what I had to do to accomplish this wasn't very DRY, as I had to re-serialize the request data in order to get to the user.

So I decided to factor that out in a manner reminicient of the `form_valid` and `form_invalid` methods of Django's generic views.

### DIFF
--- a/rest_framework_jwt/views.py
+++ b/rest_framework_jwt/views.py
@@ -51,24 +51,30 @@ class JSONWebTokenAPIView(APIView):
         kwargs['context'] = self.get_serializer_context()
         return serializer_class(*args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


ralokt:
I'm aware that this pull request decreases code coverage as currently measured.

However, this is the case because of refactoring for flexibility, not because I introduce any new behavior; I looked at the existing tests, they all seem to test views as a whole, and I don't see how I could contribute tests that cover my changes (although I'm open to suggestions!).

Also, #410 is evidence that I'm not alone in wanting something like this.

PR #419, opened 2018-01-25
Optional user based secret key

slykar:
Use `JWT_GET_USER_SECRET_KEY` only when you actually pass `user_id` in the `payload`.

This way you are able to use user-based secret keys and fallback to a default secret key when user is irrelevant.

Also, without this check, you will be getting `DoesNotExist` error when trying to generate a token for payload without `user_id`.

### DIFF
--- a/rest_framework_jwt/utils.py
+++ b/rest_framework_jwt/utils.py
@@ -21,7 +21,7 @@ def jwt_get_secret_key(payload=None):
         - password is changed
         - etc.
     """
-    if api_settings.JWT_GET_USER_SECRET_KEY:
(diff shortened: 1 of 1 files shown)


slykar:
This could be also achieved with #416 from what I can see

sergeynikiforov:
@slykar Hey, what about wrapping the call to `User.objects.get(...)` in `try...except` catching `DoesNotExist` errors and re-raising them as `jwt.InvalidTokenError` or smth?

slykar:
@sergeynikiforov I'm not sure. What I would like to achieve is to use the default key if `user_id` is not passed. I wan't to control whether to use user based secret or not when calling `jwt_encode_handler`.

PR #434, opened 2018-04-15
Add CSRF_COOKIE to prevent csrf when using JWT_AUTH_COOKIE

bmpenuelas:
To prevent Cross-Site Request Forgery when using JWT_AUTH_COOKIE, the `csrftoken` cookie will also be set when issuing the JWT authentication token.

### DIFF
--- a/rest_framework_jwt/views.py
+++ b/rest_framework_jwt/views.py
@@ -3,6 +3,8 @@ from rest_framework import status
 from rest_framework.response import Response
 from datetime import datetime
 
+from django.middleware import csrf
+
 from .settings import api_settings
 from .serializers import (
     JSONWebTokenSerializer, RefreshJSONWebTokenSerializer,
@@ -53,6 +55,9 @@ class JSONWebTokenAPIView(APIView):
 
     def post(self, request, *args, **kwargs):
         serializer = self.get_serializer(data=request.data)
+
+        if api_settings.CSRF_COOKIE:
+            csrf.get_token(request)
 
         if serializer.is_valid():
             user = serializer.object.get('user') or request.user
--- a/docs/index.md
+++ b/docs/index.md
@@ -292,6 +292,11 @@ procedure will also look into this cookie, if set. The 'Authorization' header ta
 
 Default is `None` and no cookie is set when creating tokens nor accepted when validating them.
 
+### CSRF_COOKIE
+To prevent Cross-Site Request Forgery when using JWT_AUTH_COOKIE, the `csrftoken` cookie will also be set when issuing the JWT authentication token. This works in conjuntion with django csrf middleware. The cookie contains another token which should be included in the 'X-CSRFToken' header.
+
+Default is `False`. Nevertheless, if you are using JWT_AUTH_COOKIE, it is recommended that CSRF_COOKIE is set to `True`.
+
 ## Extending `JSONWebTokenAuthentication`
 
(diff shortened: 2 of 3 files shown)


yurikulaz:
I think this changes are not enough for CSRF protection when JWT_AUTH_COOKIE option is enabled.  Your patch adds csrf cookie, but looks like it's value will be never checked, because django-rest-framework explicitly wraps every view in csrf_exempt().

https://github.com/encode/django-rest-framework/blob/master/rest_framework/views.py#L134 

I believe that BaseJSONWebTokenAuthentication authentication class should enforce csrf validation when needed. Like in rest_framework.authentication.SessionAuthentication, which explicitly call CSRFCheck?

jheld:
@bmpenuelas Thanks for writing this patch! Is this PR a comprehensive fix (with using the csrf decorator)? Is there any reason why I shouldn't make a custom `.whl` for this before merge? Is there a timeline for merge?

bmpenuelas:
Yes @jheld , using the csrf decorator this fix alone can provide the desired protection.

paul110590:
Does this protect against login csrf?

paul110590:
Hi @bmpenuelas , what I mean is the CSRF token is not set until the JWT is issued, i.e. _after_ the user has logged in. Therefore an attacker could log in via CSRF and the user would think they were logged in to their own account. The user would then unwittingly be performing actions on the attacker's account. Or is there some protection against this that I am missing?

bmpenuelas:
Protection is generally geared toward avoiding forging user requests. That protection is provided by the Anti-CSRF token and the JWT token working together. 

In the particular case you mention, at no time is the user session compromised. It's not a threat like user-impersonating CSRF, but it's true it can have privacy implications. I don't want to go off-topic here since this isn't related to the pull-request itself, but CSRFToken can protect your application from this situation too. For example, you can issue an Anti-CSRF token for the login process, and store it as both: a hidden form value, and a cookie. To validate the login request, both values must match, and since an attacker can't read them both, the login will fail.

paul110590:
It's a different type of csrf to the more common session riding, but it's still a security flaw we need to protect against.

This is where I think it has relevancy to the pull request: To protect against login-csrf we need to decorate our login view with csrf_protect. We then need to create a view that will set the csrftoken cookie, so that we can fetch it prior to login and our login request is not blocked. This being the case, we no longer need to set the csrftoken cookie when we issue the JWT, because it is already set.

Basically, if we follow the approach of protecting ourselves from login-csrf we will also automatically be protecting ourselves from 'normal' csrf (provided we guard all unsafe views with csrf_protect). Whereas if we follow the approach of protecting from 'normal' csrf, we are still leaving ourselves open to login-csrf.

bmpenuelas:
Yes, this PR already does that. CSRF tokens are issued for every request, not only after login, so you can use them for login, or to protect other requests, the same principles apply.

The problem before this PR was that when you used the JWT token for authentication, the CSRF token was ignored.

rest_framework_jwt/views.py
```diff
@@ -3,6 +3,8 @@
 from rest_framework.response import Response
 from datetime import datetime
 
+from django.middleware import csrf
```
jayvdb: just a consideration as I saw this referred to at https://github.com/Styria-Digital/django-rest-framework-jwt/issues/4 , maybe it is appropriate to only import this if `api_settings.CSRF_COOKIE` is enabled, e.g. as a delayed import within the `if` below.  

Then any site not using csrf middleware doesnt get hit with an extra module being import that isnt used.  Are there many sites doing that?

Also pep8 tools would complain bitterly about that.

pc-trent:
Any ideas when this PR is going to be merged into the package? Does it need work, can I help in any way getting this out the door?

jayvdb:
@pc-trent offer sacrifices to @gcollazo and @jpadilla . c.f. https://github.com/GetBlimp/django-rest-framework-jwt/issues/449

pc-trent:
> @pc-trent offer sacrifices to @gcollazo and @jpadilla . c.f. #449

I'll take that as a "this project is dead"

jheld:
We should hope not. Maybe we can ask for maintainer privileges? Including PyPI. We've been waiting for this fix for a year.

PR #443, opened 2018-06-01
Adding test support for Django 2.0 / DRF 3.7, 3.8

rolafium:
Hello.

I updated the test skip in `test_serializers` as it was comparing the Django version to 1.10 by checking only the second digit, which fails for Django 2.0. I also updated the tox.ini, .travis.yml and the documentation accordingly.

I removed the 3.3 tests as it fails the test (can't create the virtualenv, the bug is found here: https://github.com/tox-dev/tox/issues/644).

### DIFF
--- a/tests/test_serializers.py
+++ b/tests/test_serializers.py
@@ -15,8 +15,8 @@ User = get_user_model()
 drf2 = rest_framework.VERSION < StrictVersion('3.0.0')
 drf3 = rest_framework.VERSION >= StrictVersion('3.0.0')
 
-django_version = StrictVersion('%s.%s.%s' % django.VERSION[:3])
-django2 = DJANGO_VERSION > StrictVersion('1.10.0')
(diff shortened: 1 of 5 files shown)


rolafium:
Can I have any feedback on this pull request? Also, are you moving the project to GitLab or staying on GitHub?

tests/test_serializers.py
```diff
@@ -15,6 +15,9 @@
 drf2 = rest_framework.VERSION < StrictVersion('3.0.0')
 drf3 = rest_framework.VERSION >= StrictVersion('3.0.0')
 
+django_strict_version = StrictVersion('%s.%s.%s' % django.VERSION[:3])
+django2 = django_strict_version > StrictVersion('1.10.0')
```
jayvdb: And then Django 1.11 happened ;-)

In any case, the older PR https://github.com/GetBlimp/django-rest-framework-jwt/pull/398 did this.

PR #447, opened 2018-07-01
Update ` request.user ` after authenticating.

monokrome:
If we don't do this, then `request.user` will be the wrong instance after authenticating - which can lead to some weird issues. Specifically, in projects that use `request.user` in response middleware.


monokrome:
Anyone here? @jpadilla

monokrome:
@jpadilla this is unfortunate. No response? It has been 9 months...

PR #458, opened 2018-11-08
Have api_settings respond to settings_changed

dakotahawkins:
This is identical to what happens for `rest_framework.settings.api_settings`.

It should allow (for example) overriding the expiration deltas for a unit test.


dakotahawkins:
Eh, this doesn't seem like it's going to work anytime soon, so closing this for now.

PR #471, opened 2019-02-26
added setting JWT_PUBLIC_KEY_USING_HEADER_HANDLER

devsnd:
added setting JWT_PUBLIC_KEY_USING_HEADER_HANDLER to dynamically fetch public keys based on the header

when setting an import path to a callable into JWT_PUBLIC_KEY_USING_HEADER_HANDLER, it will be called with the JWT header as parameter. This allows to e.g. fetch dynamically changing certificates, like
google does it for example.


devsnd:
Dear DRF-JWT developers,

I made this small change that allows to get a public key based on the header of the JWT token. This allows for easy integration with e.g. firebase, as the public keys and certificates are rotated hourly.

Here is some example code of how you could use this:


    import requests
    from OpenSSL import crypto

	def get_public_key_for_payload(unverified_payload):
	    kid = unverified_payload['kid']
	    certificates = requests.get('https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com').json()
	    certificate = certificates[kid]
	    pubkey = certificate.get_pubkey()
	    pem_pubkey = crypto.dump_publickey(crypto.FILETYPE_PEM, pubkey)
	    return pem_pubkey

Then you could use this callback to get the right pubkey like this:

    JWT_PUBLIC_KEY_USING_HEADER_HANDLER = 'my_package.get_public_key_for_payload'

Please note that you should cache the request based using the `kid` as cache key in production!

devsnd:
I fixed the JWT_AUTH settings being clobbered by DRF when calling `reload` on the APISettings object. This means that `@override_settings` decorators now work properly.

By fixing the settings being clobbered by the original APISettings implementation of
restframework, we can now listen to settings changes using the `setting_changed` signal.

This signal is used by the `@override_settings` decorator. This means that you can now perform
tests that would for example override the JWT_VERIFY_EXPIRATION setting, allowing for testing
without generating new tokens on the fly.

E.g. this works now:

    @override_settings(JWT_AUTH=dict(settings.JWT_AUTH, JWT_VERIFY_EXPIRATION=False))
    def test(self):
        assert can_auth_with_expired_token()

Issue #433, opened 2018-04-14
Django 2.0


burgundy93:
Any plans for Django 2.0 support?

kamilgregorczyk:
It seems to work for me, do you have any problems?

gyukebox:
If it is compatible with django 2.0, wouldn't be better if the docs could be updated? - #435

burgundy93:
I got it working. :D Thanks guys!

budiardy:
It doesn't work well with django 2.0. 
The issue is: failed to overwrite 'JWT_RESPONSE_PAYLOAD_HANDLER'  setting.
the default is:
'JWT_RESPONSE_PAYLOAD_HANDLER': 'rest_framework_jwt.utils.jwt_response_payload_handler',
want to become:
'JWT_RESPONSE_PAYLOAD_HANDLER': 'jwt_response_payload_handler',
 i put the "def jwt_response_payload_handler" in the same page in the django settings file.
I guess it because the different version of Django

jadhavmanoj:
Any update on support django 2.x? I am now evaluating which JWT auth package use for python 3.x & Django 2.x.

krdeepak:
Same here. I tried https://github.com/davesque/django-rest-framework-simplejwt but this seems to be more widely adopted. Is someone working on django 2.0 official support?

jadhavmanoj:
@krdeepak I am used in Django 2.x. I don't see any issue.

stunaz:
@jadhavmanoj but the tests are not against Django 2.x in this repo

jadhavmanoj:
@stunaz true. currently, I am writing test cases for Django 2.1 support. I will send PR once I finished.

jadhavmanoj:
@stunaz  Please check PR #479  All test cases executed except for python version 3.3. tox has some issue for python version 3.3.

stunaz:
cool @jadhavmanoj . Now, good luck to get it merged. this project is abandoned, hasnt get updated for years ... see https://github.com/GetBlimp/django-rest-framework-jwt/issues/449

PR #479, opened 2019-04-27
Django 2.0 supprt 

jadhavmanoj:
I have changed package version to compatible with Django latest version. Also changed check for oauth2_provider import.

All existing test cases are running except for Python: 3.3 support.  getting this error `RuntimeError: Python 3.4 or later is required` for tox.  tox not working on 3.3 version.

#433


jadhavmanoj:
Should i remove testing support for python version 3.3?.  Because it is running for all other version successfully.

PR #480, opened 2019-05-21
add JWT_EXPIRATION_DELTA's judge

wnz27:
I think if exp time passed, can't execute the refresh operation~
I don't know what I think is right, whether it suitable for real application!
expect your reply and your wisdom explain!!


wnz27:
when I running, I thought it problem deeply. I think I was wrong, real application maybe when exp time passed, then we call refresh func,, I'm very sorry for disturb!

PR #489, opened 2019-08-07
Update README.rst with the new documentation URL

masoodHK:
I was going through your README file and I noticed that the documentation link is outdated and it leads to a site that doesn't exist while the documentation of the module exists in another link, i.e. https://jpadilla.github.io/django-rest-framework-jwt/. I figured it needed to have the documentation link changed to the new one so that developers who would want to use this module wouldn't have to be lost on where the documentation really is, hence why I made a pull request.


jpadilla:
Thanks

